Issue #113, opened 2018-05-24
Add a parameter type for BigNumber


JamesLefrere:
Currently, numeric values returned from the contracts are all being converted from BigNumbers to numbers; this is useful for IDs, but less accurate/potentially confusing for other numbers, such as amounts. 

Consider either adding a new parameter type for BigNumbers, or conversely, add a parameter type for numeric IDs, and leave all other numbers as BigNumbers.

This task will entail going through the Callers/Senders and changing parameter types as appropriate.

Issue #86, opened 2018-05-14
Better handling for 'empty' responses


JamesLefrere:
It's currently not clear when some Callers (e.g. `ColonyClient.GetTask`) contain an empty result. In these cases, perhaps use the incremental IDs to determine whether e.g. a task really exists (a taskId over the `ColonyClient.getTaskCount` result will not exist).

chmanie:
Was this handled by #121?

PR #121, opened 2018-05-26
Add declarative parameter type mapping for ContractMethods

JamesLefrere:
## Description

In order to add some new features (more flexible number support #113 , custom parameter types, checking against empty values #86 ), this PR makes the definition of the parameter types ('address', 'boolean', 'string', 'number') declarative. 

All of the logic to validate input values, convert input values to contract arguments, and convert contract responses to output values is now defined in one place.

It's also possible to define custom types in other packages (such as `@colony/colony-js-client`.

As far as interacting with the client is concerned, there should be no differences; the tests/integration tests should confirm this.

## Other changes

* Added much more thorough test coverage for each parameter type
* Move old imported functions (`validate, getMethodArgs, getMethodReturnValue`) to `ContractMethod` (the functionality should be the same)
* `ContractMethod`: Simplify performing validation, converting input/output values
* Add `getValidatedArgs` public method for `ContractMethod`
* Export `addParamType` from `@colony/colony-js-contract-client`
* Adds support for collecting validation error reasons
* Add jest flow-typed definition (Flow was unable to find some definitions without it)

## TODO

> ⚠️  NOTE: Please make sure all items are checked or remove the TODO list before closing the PR

- [ ] Run the colonyDapp integration tests for this branch and confirm that they pass

Contributes to #86
Contributes to #113


chmanie:
Can we run the integration tests before I start reviewing?

JamesLefrere:
Great idea, will do 👍

JamesLefrere:
When this branch is rebased on https://github.com/JoinColony/colonyJS/pull/124, and colonyDapp is running https://github.com/JoinColony/colonyDapp/pull/116, I can confirm that the integration tests pass.

packages/colony-js-contract-client/src/__tests__/ContractMethod.js
```diff
@@ -45,32 +115,75 @@ describe('ContractMethod', () => {
...
+    sandbox.spyOn(method, '_getMethodArgs');
+
+    expect(method.getValidatedArgs(inputValues)).toEqual([
+      'converted input: 1',
+    ]);
+    expect(method.validate).toHaveBeenCalledWith(inputValues);
+    expect(method._getMethodArgs).toHaveBeenCalledWith(inputValues);
+  });
+
+  // TODO test getValidatedArgs
```
chmanie: Is this still TODO?

packages/colony-js-contract-client/src/__tests__/paramTypes.js
```diff
@@ -0,0 +1,150 @@
...
+
+    // Converting output values
+    expect(convertOutputValue('0x123', 'address')).toBe('0x123');
+    expect(isValidAddress).toHaveBeenCalledWith('0x123');
+
+    isValidAddress.mockReturnValueOnce(false);
+    expect(convertOutputValue(null, 'address')).toBe(null);
+
+    // Converting input values
+    expect(convertInputValue('0x123', 'address')).toBe('0x123');
```
chmanie: Should we check the length of addresses?

packages/colony-js-contract-client/src/modules/paramTypes.js
```diff
@@ -0,0 +1,88 @@
...
+ * For each ParamType (e.g. address), defines how a value should be validated
+ * or converted for input/output.
+ */
+const PARAM_TYPE_MAP: {
+  [type: ParamTypes]: ParamTypeDef,
+} = {
+  address: {
+    validate: isValidAddress,
+    convertOutput(value) {
+      return isValidAddress(value) ? value : null;
```
chmanie: Should we really cast things to `null` if they don't match the required type? I find this a little hard.

packages/colony-js-contract-client/src/__tests__/paramTypes.js
JamesLefrere: Yes, we do check the length for the input: `isValidAddress` does this for us, e.g.`validateValue(myAddress, 'address')` calls `web3.utils.isAddress`: https://github.com/ethereum/web3.js/blob/1.0/packages/web3-utils/src/utils.js#L89

packages/colony-js-contract-client/src/modules/paramTypes.js
JamesLefrere: The idea is to clean the 'null-ish' values that we get when the data isn't set in the contracts. 

For example, `getTaskRole` will still return something for the user's address if there is no user assigned to the role for the task in question, but it's `0x00000...` etc rather than a valid address.

I think it makes sense to say 'ok, a value was returned, but we're coercing it to `null` because we know it's not valid for this type', because doing otherwise could cause some unexpected behaviour (making assumptions that the data we get back from the Callers is always valid). 

In a follow-up PR, some Callers will do some further validation if the results they get back are like this (a suggestion from Alex) – we would e.g. call `getTaskCount` and compare it with the task ID, to see if the task actually exists.

chmanie: Ah ok. All of that makes sense. Is there an issue for the follow up PR? If not, could you create one?

packages/colony-js-contract-client/src/__tests__/ContractMethod.js
JamesLefrere: Sorted now!

PR #122, opened 2018-05-26
Add 'bignumber' parameter type

### LINKED ISSUE #113: Add a parameter type for BigNumber

JamesLefrere:
## Description

Adds a new parameter type 'bignumber' and implements this in the client for Callers/Senders that use amounts.

The idea is that the 'number' type is good for IDs and integers (it automatically converts BigNumbers from the contract responses to numbers), but less useful for amounts, where we need accuracy.

* Fixes `isBigNumber` function's handling of null values
* Will affect the documentation!

Resolves #113

### DIFF
--- a/docs/_API_ColonyClient.md
+++ b/docs/_API_ColonyClient.md
@@ -146,7 +146,7 @@ A promise which resolves to an object containing the following properties:
 
 |Return value|Type|Description|
 |---|---|---|
-|amount|number|Amount of specified tokens to payout for that task and a role|
+|amount|BigNumber|Amount of specified tokens to payout for that task and a role|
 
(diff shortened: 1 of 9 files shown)


JamesLefrere:
Note to self: don’t forget to check the docs!

PR #123, opened 2018-05-26
Add 'role' custom parameter type

JamesLefrere:
## Description

Adds a 'role' custom parameter type within `@colony/colony-js-client`. 

Instead of calling e.g.:

```js
await colonyClient.setTaskRoleUser.send({ taskId: 1, role: 2, user: '0x...' });
```

We can now call:

```js
await colonyClient.setTaskRoleUser.send({ taskId: 1, role: WORKER_ROLE, user: '0x...' });
```

or even just:

```
await colonyClient.setTaskRoleUser.send({ taskId: 1, role: 'WORKER', user: '0x...' });
```

Additionally, if a Caller were to return a role, it would also be converted from a BigNumber/number to a string:

```
const { role } = await colonyClient.callerThatReturnsRole.call({ taskId: 1 });
// role -> 'WORKER'
```


chmanie:
Note to self: don’t forget to check the docs

PR #126, opened 2018-05-28
Better handling for 'empty' responses

### LINKED ISSUE #86: Better handling for 'empty' responses

JamesLefrere:
## Description

This PR improves the handling of 'empty' responses (responses from the contract functions called by `Caller`s which have null values). 

If some of the values returned by a `Caller` are null, and that `Caller` has a `validateEmpty` function, the call may throw an error if the validation deemed the result to be empty.

For example, when we call

```js
const task = await ColonyClient.getTask.call({ taskId: 3 });
```

If task 3 doesn't exist, we should get some null values in the response (e.g. `specificationHash: null`). The `validateEmpty` function makes a call to `ColonyClient.getTaskCount` and compares the task count with the ID we wanted; if the ID is greater than the count, we know for sure that the task doesn't exist, and an error is thrown ('Task ID 3 not found').
  

## Other changes

* Add missing `getDomain` Caller to `ColonyClient`, with `validateEmpty` based on `getDomainCount`
* Add `validateEmpty` function for task-related Callers, `getDomain` and `getSkill`
* Check for empty/invalid responses in `ColonyNetworkClient.getColonyClient`/`ColonyNetworkClient.getMetaColonyClient`
* Rename `getSkill` parameter to `skillId` (was `id`) for consistency
* Update docs
* Bugfix: fixed a bug where the `isBigNumber` utility function didn't look for the correct property in ethers-ified BigNumbers.

## TODOs

- [ ] Confirm that the integration tests work

Resolves #86

### DIFF
--- a/packages/colony-js-utils/src/isBigNumber.js
+++ b/packages/colony-js-utils/src/isBigNumber.js
@@ -8,7 +8,7 @@ export default function isBigNumber(value: any) {
     // XXX Some libraries (cough *ethers* cough) wrap BigNumbers in a way
     // that breaks `isBN`; this is a workaround for that issue:
     (value != null &&
-      Object.hasOwnProperty.call(value, 'bn') &&
+      Object.hasOwnProperty.call(value, '_bn') &&
       // eslint-disable-next-line no-underscore-dangle
       BigNumber.isBN(value._bn))
   );
--- a/docs/_API_ColonyClient.md
+++ b/docs/_API_ColonyClient.md
@@ -44,6 +44,25 @@ A promise which resolves to an object containing the following properties:
 |---|---|---|
 |secret|string|keccak256 hash of joint Salt and Value|
 
+### `getDomain.call({ domainId })`
+
+Gets the selected domain's local skill ID and funding pot ID
+
+**Arguments**
+
+|Argument|Type|Description|
+|---|---|---|
+|domainId|number|ID of the domain|
+
+**Returns**
+
+A promise which resolves to an object containing the following properties:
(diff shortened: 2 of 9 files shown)


JamesLefrere:
Added a TODO, will do 👍

JamesLefrere:
screen shot 2018-05-28 at 22 44 17

☝️ confirming that this passes tests (when `colonyDapp` is on branch `maintenance/support-colonynetworkclient-updates`).

packages/colony-js-utils/src/isBigNumber.js
```diff
@@ -8,7 +8,7 @@ export default function isBigNumber(value: any) {
     // XXX Some libraries (cough *ethers* cough) wrap BigNumbers in a way
     // that breaks `isBN`; this is a workaround for that issue:
     (value != null &&
-      Object.hasOwnProperty.call(value, 'bn') &&
+      Object.hasOwnProperty.call(value, '_bn') &&
```
chmanie: Was this a bug?

JamesLefrere: It was, I'll put it in the description

JamesLefrere: PS nice gif 😆

PR #131, opened 2018-05-28
Autogenerate docs when commiting

### LINKED ISSUE #91: Automatically generate API docs using CI and a template file
Let’s say colony-js gets updated. We put a placeholder into a template file in `/docs` (which is pretty much what we have right now except it doesn't contain the generated documentation and doesn’t begin with an underscore).

When building colony-js on the CI we

1) Run the docgen
2) Place the contents into the template file (where the placeholder is)
3) Generate the underscored file from that
4) Profit

Additionally add:

- [ ] Types table in the Contract client docs (ColonyClient, ColonyNetworkClient)

chmanie:
This adds a command to the package.json of the colony-client: build:docs. This will run the doc generation process of the contract client classes (currently ColonyClient and ColonyNetworkClient). This uses template files to generate the actual doc files (these sit in the docs folder of the colony client).

Also, it will trigger this build step pre every commit (it's kinda unnecessary but it only takes 0.5s so for now that's probably acceptable).

Closes #91.

### DIFF
--- a/packages/colony-js-client/docs/_API_ColonyClient.template.md
+++ b/packages/colony-js-client/docs/_API_ColonyClient.template.md
@@ -1,0 +1,21 @@
+---
+title: ColonyClient
+section: API
+order: 3
+---
+
+The `ColonyClient` class is a standard interface for interactions with the on-chain functions and events described in `IColony.sol`
+
+These interactions are generally concerned with functions and events internal to a colony, such as creating a task, assigning a work rating, or moving funds between pots.
+
+For functions and events that concern the colonyNetwork as a whole, refer to the [ColonyNetworkClient API](/colonyjs/api-colonynetworkclient/)
+
+==TOC==
+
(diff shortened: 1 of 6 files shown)


packages/colony-js-client/scripts/docgen.js
```diff
@@ -5,6 +5,19 @@ const path = require('path');
 const parser = require('flow-parser');
 const types = require('ast-types');
 
+const CONTRACT_CLIENTS = [
```
JamesLefrere: nice one!

packages/colony-js-client/docs/_API_ColonyClient.template.md
```diff
@@ -0,0 +1,21 @@
+---
```
JamesLefrere: Hooray for templates!

chmanie:
Rrrrrrebased!

PR #135, opened 2018-05-29
Bug fix: Convert non-hex string input values to hex

JamesLefrere:
## Description

The parameter mapping introduce in #121 contained a bug relating to string conversion: the statement was the wrong way round, leading to non-hex strings not being converted to hex strings. 

The tests have been made more explicit in order to try and prevent this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v.NEXT
 
+**Bug fixes**
+
+* Fixed the documentation site link in the `README`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## v.NEXT
 
+**Bug fixes**
+
+* Fixed the documentation site link in the `README`
```
JamesLefrere: NB this refers to #134

JamesLefrere:
Where do you get these gifs @chmanie  😄

PR #158, opened 2018-06-03
Add IPFS hash parameter type

JamesLefrere:
## Description (Required)

This PR adds a new `ipfsHash` parameter type, to more easily store/retrieve IPFS hashes from the contracts. 

It also implements the `ipfsHash` type in these `ColonyClient` methods: `getTask`, `createTask`, `submitTaskDeliverable`, and `setTaskBrief`.

* Ensure that the returned values in `ColonyClient.getTask` are handled properly
* Label optional (maybe-null) types in the docgen script

## TODO

> ⚠️  NOTE: Please make sure all items are checked or remove the TODO list before closing the PR

- [x] Test with the integration testing

## Deps

**New dependencies**:

- `bs58` : Encode/decode base 58


packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -743,15 +744,15 @@ export default class ColonyClient extends ContractClient {
         multisigFunctionName: 'executeTaskChange',
         nonceFunctionName: 'getTaskChangeNonce',
       });
-    makeExecuteTaskChange('setTaskBrief', [['specificationHash', 'string']]);
+    makeExecuteTaskChange('setTaskBrief', [['specificationHash', 'ipfsHash']]);
```
chmanie: Should we be consistent with the casing here?

ipfsHash vs. bignumber

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -743,15 +744,15 @@ export default class ColonyClient extends ContractClient {
...
     makeExecuteTaskChange('setTaskDueDate', [['dueDate', 'number']]);
     makeExecuteTaskChange('setTaskWorkerPayout', [
       ['token', 'address'],
       ['amount', 'bignumber'],
     ]);
     makeExecuteTaskChange('setTaskManagerPayout', [
       ['token', 'address'],
-      ['amount', 'nubignumbermber'],
+      ['amount', 'bignumber'],
```
chmanie: 🙌

packages/colony-js-client/src/ColonyClient/index.js
JamesLefrere: Yes, camelCase is probably more future-proof; I've done a case-sensitive find and replace, seems to do the trick.

PR #161, opened 2018-06-04
Define default values for Caller/Sender params

### LINKED ISSUE #153: Better domain handling in createTask sender
Currently the `createTask` sender requires a `domainId` to be set. We want to make it a little bit easier for people to use who don't want to concern themselves with domains. So it'd be good to *set the default value of `domainId` to 1* (which is the root domain).

Also, when specifying a `domainId` which doesn't exist, we get a revert error. We should add some async validation so that it checks whether the domain exists before trying to create the task.

TODO:

- [x] Make 1 the default domain
- [x] Adjust the docs (also for the createTask sender) (@gichiba ?)
- [ ] Validate that the domain exists before trying to create a task

JamesLefrere:
## Description

This PR allows developers to define default values for Caller/Sender parameters, and implements this for `ColonyClient.createTask`.

* Add a third (optional) parameter for the `Param` flow type; the default value
* Fall back to default values in `ContractClient._parseInputValues`
* Add a default value (1) for `domainId` in `ColonyClient.createTask`

Contributes to #153

### DIFF
--- a/packages/colony-js-client/src/ColonyClient/index.js
+++ b/packages/colony-js-client/src/ColonyClient/index.js
@@ -11,7 +11,13 @@ import type { ContractClientConstructorArgs } from '@colony/colony-js-contract-c
 
 import ColonyNetworkClient from '../ColonyNetworkClient/index';
 import GetTask from './callers/GetTask';
-import { ROLES, WORKER_ROLE, EVALUATOR_ROLE, MANAGER_ROLE } from '../constants';
+import {
+  ROLES,
+  WORKER_ROLE,
+  EVALUATOR_ROLE,
+  MANAGER_ROLE,
+  DEFAULT_DOMAIN_ID,
+} from '../constants';
 
 type Address = string;
 type Role = $Keys<typeof ROLES>;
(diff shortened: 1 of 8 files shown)


packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -663,7 +669,10 @@ export default class ColonyClient extends ContractClient {
     });
     this.addSender('createTask', {
       functionName: 'makeTask',
-      input: [['specificationHash', 'ipfsHash'], ['domainId', 'number']],
+      input: [
+        ['specificationHash', 'ipfsHash'],
+        ['domainId', 'number', DEFAULT_DOMAIN_ID],
```
chmanie: 👍

packages/colony-js-contract-client/src/classes/ContractMethod.js
```diff
@@ -4,7 +4,7 @@
 import isPlainObject from 'lodash.isplainobject';
 import { makeAssert } from '@colony/colony-js-utils';
 
-import type { ContractMethodArgs, ParamTypePairs } from '../flowtypes';
+import type { ContractMethodArgs, Params } from '../flowtypes';
```
chmanie: I like the renaming there.

PR #169, opened 2018-06-05
Additional small changes to docs

gichiba:

## Description (Required)
Minor changes to documentation to fix before re-deploying docs at Hackathon Kickoff


gichiba:
I don't think there's a reason we can't just merge these now and publish them.

PR #190, opened 2018-06-17
Wait for in-progress transactions to be mined before finding the receipt

JamesLefrere:
## Description

This PR makes the Senders wait for in-progress transactions to be fully confirmed before getting the receipt. I noticed this issue when testing with Rinkeby; with a low gas price, the receipt was sometimes null; now, it will wait until the timeout runs out, which has also been changed: 60 minutes for mainnet, 5 minutes otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## v.NEXT
 
+**Enhancements**
+
+* Wait for in-progress transactions when getting the transaction receipt (`@colony/colony-js-contract-client`, `@colony/colony-js-adapter-ethers`)
 
 ## v1.4.1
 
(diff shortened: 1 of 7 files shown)


chmanie:
It seems on mainnet this timeout has to be significantly higher ~60min.

JamesLefrere:
Good idea. I've now changed it such that it sets a 60 minute timeout if the client is using mainnet, and otherwise 5 minutes.

JamesLefrere:
I guess we'll go with this for the time being (we need the fix), but it's worth exploring more thorough solutions, potentially based on estimating gas cost/getting gas price/selecting gas price, then estimating a reasonable timeout (à la Peepeth).

PR #193, opened 2018-06-19
Improve consistency of token attribute

chmanie:

_

Doc fixes. For consistency and better naming we renamed `source` to `tokenAddress`,
moved the payable address type to the contract client, and renamed it to `tokenAddress`.
_


chmanie:
@gichiba @JamesLefrere what do you think?

PR #224, opened 2018-06-26
Add missing unit tests for colony-js-utils

sprusr:
## Description

Adds missing unit tests for `colony-js-utils`, plus fixes for `raceAgainstTimeout` tests.

### DIFF
--- a/packages/colony-js-utils/jest.conf.json
+++ b/packages/colony-js-utils/jest.conf.json
@@ -1,6 +1,6 @@
 {
     "rootDir": "src",
-    "collectCoverageFrom" : ["*.{js}", "**/*.{js}", "!index.js"],
+    "collectCoverageFrom" : ["*.{js}", "**/*.{js}"],
     "coverageDirectory": "../coverage",
     "coverageThreshold": {
         "global": {
--- a/packages/colony-js-utils/jest.conf.json
+++ b/packages/colony-js-utils/jest.conf.json
@@ -1,6 +1,6 @@
 {
     "rootDir": "src",
-    "collectCoverageFrom" : ["*.{js}", "**/*.{js}"],
+    "collectCoverageFrom" : ["*.{js}", "**/*.{js}", "!index.js"],
     "coverageDirectory": "../coverage",
     "coverageThreshold": {
         "global": {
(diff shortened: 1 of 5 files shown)


packages/colony-js-utils/jest.conf.json
```diff
@@ -1,6 +1,6 @@
 {
     "rootDir": "src",
-    "collectCoverageFrom" : ["*.{js}", "**/*.{js}"],
+    "collectCoverageFrom" : ["*.{js}", "**/*.{js}", "!index.js"],
```
chmanie: Haha, you couldn't leave it be, could you?
I actually don't think that's a good idea as we might add some more index files in the future and also might add some logic to it. This will basically just ignore those files from coverage. Let's not do this.

packages/colony-js-utils/src/__tests__/isBigNumber.js
```diff
@@ -0,0 +1,27 @@
...
+
+import isBigNumber from '../isBigNumber';
+
+jest.mock('bn.js');
+
+describe('isBigNumber', () => {
+  test('Null value', () => {
+    BigNumber.isBN.mockReturnValueOnce(false);
+
+    expect(isBigNumber(null)).toBe(false);
```
chmanie: When mocking `bn.js` like this you should actually check whether the internal function was called correctly.

packages/colony-js-utils/src/__tests__/isBigNumber.js
```diff
@@ -0,0 +1,27 @@
...
+
+  test('Ethers wrapped', () => {
+    BigNumber.isBN.mockReturnValueOnce(false).mockReturnValueOnce(true);
+
+    const bn = 'big number';
+    const wrappedBN = {
+      _bn: bn,
+    };
+
+    expect(isBigNumber(wrappedBN)).toBe(true);
```
chmanie: Same as above. You're actually doing it in the next file.

packages/colony-js-utils/jest.conf.json
JamesLefrere: I sent over the suggestion, sorry! 😉

sprusr: Couldn't resist the green! But yeah good point, will remove this

PR #226, opened 2018-06-26
Maintenance/colony js contract loader tests

sprusr:
## Description

Adds unit tests for `colony-js-contract-loader`, along with the package scripts to run them.

### DIFF
--- a/packages/colony-js-contract-loader/src/__tests__/ContractLoader.js
+++ b/packages/colony-js-contract-loader/src/__tests__/ContractLoader.js
@@ -1,0 +1,159 @@
+/* eslint-env jest */
+
+import assert from 'assert';
+import createSandbox from 'jest-sandbox';
+import ContractLoader from '../ContractLoader';
+import { DEFAULT_REQUIRED_CONTRACT_PROPS } from '../defaults';
+
+jest.mock('assert');
+
+describe('ContractLoader', () => {
+  const sandbox = createSandbox();
+
+  beforeEach(() => {
+    sandbox.clear();
+    assert.mockReset();
+  });
+
+  test('Default transform', async () => {
+    expect(
+      ContractLoader.defaultTransform()('input', 'query', 'required props'),
+    ).toBe('input');
+  });
+
+  test('Validate contract definition', async () => {
+    const contractDef = {
+      address: 'contract address',
+      abi: 'contract abi',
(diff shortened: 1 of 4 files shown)


packages/colony-js-contract-loader/src/__tests__/ContractLoader.js
```diff
@@ -0,0 +1,159 @@
...
+  const sandbox = createSandbox();
+
+  beforeEach(() => {
+    sandbox.clear();
+    assert.mockReset();
+  });
+
+  test('Default transform', async () => {
+    expect(
+      ContractLoader.defaultTransform()('input', 'query', 'required props'),
```
chmanie: I probably would have used an object instead of a string to show the actual use case but it does the job :)

packages/colony-js-contract-loader/src/__tests__/ContractLoader.js
```diff
@@ -0,0 +1,159 @@
...
+      throw new Error();
+    });
+    expect(() => {
+      // eslint-disable-next-line no-new
+      new ContractLoader({ transform: 'not a function' });
+    }).toThrow();
+    expect(assert).toHaveBeenCalledWith(false, expect.any(String));
+  });
+
+  test('_load', async () => {
```
chmanie: Maybe be a bit more descriptive here? Why are we expecting `_load()` to throw?

packages/colony-js-contract-loader/src/__tests__/ContractLoader.js
```diff
@@ -0,0 +1,159 @@
...
+    const loader = new ContractLoader();
+    // eslint-disable-next-line no-underscore-dangle
+    expect(loader._load()).rejects.toThrow();
+  });
+
+  test('Load with no query', async () => {
```
chmanie: Same for this here!

packages/colony-js-contract-loader/src/__tests__/ContractLoader.js
sprusr: I decided to keep the input the same, as (although probably json) it could conceivably be a string also, and is easy to read this way. Got rid of the other arguments though, they definitely wouldn't ever be strings and aren't required for this test anyway.

sprusr:
@chmanie thanks for feedback on this and the other PR 🙏 changes made!

PR #227, opened 2018-06-26
Improve test coverage for colony-js-contract-loader-fs

chmanie:

Does exactly what it says on the tin. 100% coverage of branches is what we have
now here.


chmanie:
@sprusr your turn!

PR #229, opened 2018-06-27
Pad input hex strings (`tokenAddress` param type)

### LINKED ISSUE #213: `0x0` is not a valid address in ethers
While the `ContractClient`'s validation allows for `0x0` to be a valid (empty) address, `ethers` doesn't allow that. Hence we should pad the `0x0` input in all methods to an address with a proper length before passing it into ethers.

Another option would be to adjust the docs everywhere to use `EMPTY_ADDRESS` (could be aliased to `ETHER`) instead of `0x0`.

JamesLefrere:
## Description

Partial empty hex strings (e.g. `0x0`) are now padded to full-length, which resolves an issue with `EthersAdapter`.

* The `isEmptyHexString` utility function now evaluates the input type
* Adds test coverage for `isEmptyHexString`
* Adds test coverage for the `tokenAddress` param type

Resolves #213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## v.NEXT
 
+**Bug fixes**
+
+* Partial empty hex strings (e.g. `0x0`) are now padded to full-length, which resolves an issue with `EthersAdapter` (`@colony/colony-js-contract-client`)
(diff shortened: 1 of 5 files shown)


JamesLefrere:
No worries :)

PR #230, opened 2018-06-28
Change `getUserRole` to `getUserRoles`

### LINKED ISSUE #200: AuthorityClient always reports `null` for a role using `getUserRole`
When running `authorityClient.getUserRole.call({ user: address }), it always reports `null` for the `role`. Sanity checking using `hasUserRole` reveals that the role is indeed different from `null`.

Digging into the code a bit shows that role numbers are `bytes32` that arrive in hex format that isn't parsed in any way here: 

https://github.com/JoinColony/colonyJS/blob/master/packages/colony-js-client/src/paramTypes.js#L19

No role: `0x0000000000000000000000000000000000000000000000000000000000000000`
Admin role: `0x0000000000000000000000000000000000000000000000000000000000000002`

ryanchristo:
## Description

Changes the `getUserRole` caller function to `getUserRoles`, which returns an array of authority roles associated with the given user address.

## TODO

- [ ] Add unit test for `getUserRoles`
- [x] Update documentation for colonyClient

Resolves #200


ryanchristo:
Closing in favor of #307

PR #232, opened 2018-07-02
Add `waitForTransactionReceipt` abstraction to adapters

### LINKED ISSUE #195: Create adapter abstraction for getTransactionReceipt
We might need [this logic](https://github.com/JoinColony/colonyJS/commit/02bd502f7e4caec6919138a524ee90cb54b42eda#diff-eda2497d32babcc12083d637fbc0b737R147) more than once (e.g. when deploying contracts). Hence it might make sense to put this into another abstraction which can live in the adapter.

JamesLefrere:
## Description

This PR adds a method on the adapters: `waitForTransactionReceipt`. This method will wait for an in-progress transaction to be mined (if necessary) before getting the receipt. The new method is now used in the Senders and also during token contract creation.

* Raises `EthersAdapter` test coverage to 100%
* Fixes a possible object destructuring issue with `Adapter.getContract` found during testing

Resolves #195

### DIFF
--- a/packages/colony-js-adapter-ethers/src/EthersAdapter.js
+++ b/packages/colony-js-adapter-ethers/src/EthersAdapter.js
@@ -54,11 +54,12 @@ export default class EthersAdapter implements IAdapter {
     this.wallet = wallet;
   }
   async getContract(query: Query): Promise<IContract> {
-    const { address, abi } = await this.loader.load(query, {
-      abi: true,
-      address: true,
-      bytecode: false,
-    });
+    const { address, abi } =
+      (await this.loader.load(query, {
+        abi: true,
+        address: true,
+        bytecode: false,
+      })) || {};
 
     if (typeof address !== 'string')
       throw new Error('Unable to parse contract address');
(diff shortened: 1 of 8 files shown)


packages/colony-js-adapter-ethers/src/EthersAdapter.js
```diff
@@ -106,11 +107,33 @@ export default class EthersAdapter implements IAdapter {
...
+    if (receipt == null)
+      throw new Error(
+        `Transaction receipt not found (transaction: ${transactionHash})`,
+      );
+    return receipt;
   }
   async waitForTransaction(transactionHash: string) {
     return this.provider.waitForTransaction(transactionHash);
   }
+  async waitForTransactionReceipt(transactionHash: string) {
```
chmanie: Two things here:
I'm a bit confused about the name here. `getTransactionReceipt` and `waitForTransactionReceipt` both return the receipt. Which one am I supposed to use? I don't have a better idea for naming the function which basically is `waitForAndGetTransactionReceipt`. Can't we just have one function? Sorry for these unstructured sentences, I'm thinking out loud here...

Second of all: could we have a custom timeout?

packages/colony-js-adapter-ethers/src/__tests__/EthersAdapter.test.js
```diff
@@ -253,4 +295,66 @@ describe('EthersAdapter', () => {
...
+    try {
+      await adapter.waitForTransactionReceipt(transactionHash);
+      expect(false).toBe(true); // Should be unreachable
+    } catch (error) {
+      expect(error.toString()).toMatch('Kaboom!');
+    }
+    sandbox.restore(adapter, 'getTransactionReceipt');
+  });
+
+  test('Signing a message', async () => {
```
chmanie: 🎉 test coverage!

packages/colony-js-adapter-ethers/src/EthersAdapter.js
JamesLefrere: Agreed. Let's go with the public method `getTransactionReceipt`, which calls `_getTransactionReceipt` and does the waiting if necessary.

Good idea on the timeout argument, added 👍

PR #233, opened 2018-07-03
Check domain ID when creating task

### LINKED ISSUE #153: Better domain handling in createTask sender

JamesLefrere:
## Description

This PR adds a `CreateTask` Sender, which performs some async validation before committing to a `send`; the given domain ID must exist on the colony, and an error will be thrown if it does not.

It also contains some refactoring to simplify parameter conversion and default values (which Flow previously highlighted as problematic).

* Split up the flow types into separate files because it was getting crazy
* Remove the third parameter (default value) for the `Param` type. This was done because Flow was complaining about how the input values for the Sender conflicted with the typings for `Params` and `EventParams`, which was itself a compromise. I felt this was getting out of hand, so the whole thing has been made simpler by specifying the default value separately from these tuples, and then applying the default values elsewhere.
* Since `Param` no longer contains a default value, don't apply default values when converting input values
* Specify an optional `defaultValues` property for `ContractMethod`
* Apply the default values in `validate` and `getValidatedArgs` with `_applyDefaultValues`
* Remove `_getMethodArgs` method
* `convertInputValues` no longer applies default values
* The `EventCallback` type might be used directly to return something; loosen the typing.

## Deps

**New dev dependencies**:

- `flow-bin` : upgraded to 0.75.0
- `flow-parser` : upgraded to 0.75.0

Resolves #153


packages/colony-js-client/src/ColonyClient/senders/CreateTask.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+type OutputValues = {
+  id: number,
+};
+
+// XXX This is a good use-case for some kind of async validation step,
+// but since the underlying method functionality is due to change very soon,
+// we're opting to not make big changes to the Sender behaviour, and simply
+// extend the `send` method to perform this async validation.
```
chmanie: 👍

packages/colony-js-contract-client/src/classes/ContractMethod.js
```diff
@@ -114,10 +150,19 @@ export default class ContractMethod<
...
 
   /**
-   * Given input values, validate them and return parsed method args.
+   * Given input values, get default values (if params are given), then
+   * validate them and return parsed method args.
    */
-  getValidatedArgs(inputValues?: any, params: Params = this.input) {
-    this.validate(inputValues, params);
-    return this._getMethodArgs(inputValues, params);
+  getValidatedArgs(
```
chmanie: This function is (except for one line) the same as the one above. Room for a refactor?

chmanie: I also didn't really get the difference immediately. Maybe also make it a bit more clear?

packages/colony-js-contract-client/src/flowtypes/index.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+
+export type {
```
chmanie: Jau!

packages/colony-js-contract-client/src/classes/ContractMethod.js
JamesLefrere: Hopefully this is a bit clearer now:

* Removed _getMethodArgs method
* Improved typing of _applyDefaultValues (it now returns InputValues)
* _applyDefaultValues can now be called without params (simply returns a copy of the input)

PR #234, opened 2018-07-04
Add event processors for new task change events

fallafeljan:
## Description

We recently introduced the `TaskDeliverableSubmitted` and `TaskWorkRatingRevealed` events on `colonyNetwork` (https://github.com/JoinColony/colonyNetwork/pull/263). This commit adds the respective event processors for these to `ColonyClient`.

### DIFF
--- a/packages/colony-js-client/src/ColonyClient/index.js
+++ b/packages/colony-js-client/src/ColonyClient/index.js
@@ -748,6 +748,16 @@ export default class ColonyClient extends ContractClient {
       ['token', 'tokenAddress'],
       ['amount', 'number'],
     ]);
+    this.addEvent('TaskDeliverableSubmitted', [
+      ['id', 'number'],
+      ['deliverableHash', 'ipfsHash'],
+    ]);
+    this.addEvent('TaskWorkRatingRevealed', [
+      ['id', 'number'],
+      // $FlowFixMe
+      ['role', 'role'],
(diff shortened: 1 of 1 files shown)


packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -748,6 +748,16 @@ export default class ColonyClient extends ContractClient {
       ['token', 'tokenAddress'],
       ['amount', 'number'],
     ]);
+    this.addEvent('TaskDeliverableSubmitted', [
+      ['id', 'number'],
+      ['deliverableHash', 'ipfsHash'],
+    ]);
+    this.addEvent('TaskWorkRatingRevealed', [
+      ['id', 'number'],
+      // $FlowFixMe
```
sprusr: Grrr this is so annoying!! Do you think it's worth removing such strict typing of `ParamTypes`? Especially as a user of `ContractClient` can add whatever types they want. I guess this is a discussion for another time though!

JamesLefrere: Yep we discussed this; I'd also be fine with changing the type to `[string, string]`.

fallafeljan: Yip, I'd thought leaving the `$FlowFixMe` would allow for easy searching once we're getting on this.

PR #236, opened 2018-07-05
Fix wrong role conceptions in documentation

fallafeljan:
## Description
`role` specifies the role that gets rated/revealed, not the role that submits/reveals the rating — it was wrongly stated before. The evaluator, however, can be specified via `setTaskRoleUser`. The user who submits/reveals her or his rating for the worker needs to be specified in that role.


gichiba:
I don't know if my commit is necessary or not, but I think that the changes made in _API_ColonyClient will be overwritten if they're not also changed in the client code comments. Also, my hooks are broken so I had to commit with `--no-verify` -- I hope that's ok...

fallafeljan:
Good point, I didn’t think about the generation!

chmanie:
Oops. Didn't catch that. Thanks @gichiba

chmanie:
Also @fallafeljan that means that you also didn't run the precommit hooks (as the docs should be built on commit).

gichiba:
Ah, this probably means that "Improve this doc" isn't really that effective for any of the API docs huh?

chmanie:
Yeah, good point, Griffin. But for that we can probably just do it like this time. It’s an easy way to point out errors, just like Jan did.

fallafeljan:
Couldn't we move (or: expand) generation of documentation to CI?

chmanie:
> Couldn't we move (or: expand) generation of documentation to CI?

Yes that certainly is possible. Want to open an issue in the backlog?

fallafeljan:
This PR is becoming somewhat stale, shall we merge it? I can't do it by myself.

chmanie:
Oh, yeah sure. If you rebase it, I'll merge it!

PR #240, opened 2018-07-13
Return isGlobalSkill in ColonyNetworkClient.getSkill

### LINKED ISSUE #239: Support `isGlobalSkill` flag from ColonyNetwork.getSkill
A `isGlobalSkill` boolean flag will be added as the third return value of `ColonyNetwork.getSkill` so we need to support this new return value.

https://github.com/JoinColony/colonyNetwork/issues/249

KevinLiLu:
## Description
Support new `isGlobalSkill` boolean flag from `ColonyNetwork.getSkill` (https://github.com/JoinColony/colonyNetwork/issues/249)

Resolves #239


JamesLefrere:
Thanks for PR Kevin, this is bundled in #252 👍

PR #273, opened 2018-09-27
Parse `SkillAdded` event properly when emitted via `ColonyClient`

JamesLefrere:
## Description

This PR ensures that the `SkillAdded` event is parsed correctly when logged in transactions from `ColonyClient` instances (previously it was only available when emitted via `ColonyNetworkClient`.

Also removes some unnecessary `await` statements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## v.NEXT
 
+**Bug fixes**
+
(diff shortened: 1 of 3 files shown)


rdig:
Please add an estimate to this

JamesLefrere:
Good point, done.

PR #278, opened 2018-10-10
Add ENS functionality

### LINKED ISSUE #277: Add support for ENS methods
`colonyNetworkENS` contains several new public methods to assist in returning an OrbitDB address. These should be supported in colonyJS so that users can retrieve the DDB address associated with their registered ENS names.

Please see [relevant new methods](https://github.com/JoinColony/colonyNetwork/commit/102ec5ec9b36645a2d6893fe6b775c155a226a6f). 

The new public methods appear to be:
- [ ] contracts/ColonyNetworkENS.sol, `getProfileDBAddress`
- [ ] contracts/ColonyNetworkENS.sol, `lookupRegisteredENSDomain`
- [ ] contracts/ColonyNetworkENS.sol, `supportsInterface`
- [ ] contracts/ColonyNetworkENS.sol, `addr `

coyotespike:
## Description

This PR updates the network and colony clients so that they support recent changes to ENS-related functionality in the contracts.

Events:

* `ColonyLabelRegistered` added
* `UserLabelRegistered` added

Callers: 

* `ColonyClient.registerColonyLabel` now takes a `colonyName` parameter
* New: `ColonyNetworkClient.getProfileDBAddress`
* New: `ColonyNetworkClient.lookupRegisteredENSDomain`
* New: `ColonyNetworkClient.getAddressForENSHash`
* New: `ColonyNetworkClient.ensSupportsInterface`

Senders:

* `ColonyNetworkClient.registerUserLabel` now takes `{ username, orbitDBPath }` parameters and emits a `UserLabelRegistered` event.

These all take or return `hexString`, `string`, `boolean`, or `address`, and so this PR defines no new types.

Closes #277.

### DIFF
--- a/packages/colony-js-client/src/ColonyClient/index.js
+++ b/packages/colony-js-client/src/ColonyClient/index.js
@@ -91,6 +91,10 @@ type TaskCanceled = ContractClient.Event<{
 type RewardPayoutCycleStarted = ContractClient.Event<{
   payoutId: number, // The payout ID logged when a new reward payout cycle has started.
 }>;
+type ColonyLabelRegistered = ContractClient.Event<{
+  colony: Address, // Address of the colony that registered a label
+  label: string, // The label registered
+}>;
 
 export default class ColonyClient extends ContractClient {
   networkClient: ColonyNetworkClient;
@@ -397,9 +401,9 @@ export default class ColonyClient extends ContractClient {
   */
   registerColonyLabel: ColonyClient.Sender<
     {
-      subnode: string, // The keccak256 hash of the label to register
+      colonyName: string, // The label to register
     },
-    {},
+    { ColonyLabelRegistered: ColonyLabelRegistered },
     ColonyClient,
   >;
   /*
@@ -771,6 +775,7 @@ export default class ColonyClient extends ContractClient {
   >;
 
   events: {
+    ColonyLabelRegistered: ColonyLabelRegistered,
     DomainAdded: DomainAdded,
     PotAdded: PotAdded,
     RewardPayoutCycleStarted: RewardPayoutCycleStarted,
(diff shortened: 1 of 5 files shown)


coyotespike:
And, as we just discussed, the events `UserLabelRegistered`, `ColonyLabelRegistered`.

Also just to note, all the functions above are `public`, except `supportsInterface` which is `external`.

chmanie:
Cool! May I know whether the senders are going to be there as well?

coyotespike:
Do you mean `registerColonyLabel`, `registerUserLabel`, and `setUpRegistrar`? Those are already in colonyJS, if so.

packages/colony-js-client/src/ColonyNetworkClient/index.js
```diff
@@ -172,6 +183,54 @@ export default class ColonyNetworkClient extends ContractClient {
     },
     ColonyNetworkClient,
   >;
+  /*
+      Returns the address of a colony when given the hashed ENS username
```
coyotespike: This description doesn't look quite right. It should be something like `Returns the database address of a user when given the hashed ENS username`

packages/colony-js-client/src/ColonyNetworkClient/index.js
```diff
@@ -172,6 +183,54 @@ export default class ColonyNetworkClient extends ContractClient {
...
+    */
+  getProfileDBAddress: ColonyNetworkClient.Caller<
+    {
+      nameHash: HexString, // The hashed human-readable ENS name
+    },
+    {
+      orbitDBAddress: Address, // Address of the UserProfile DDB
```
coyotespike: Note that an `orbitDBAddress` looks like `/orbitdb/IPFSHash/myDbName`. To me type `address` implies an Ethereum address.

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -91,6 +91,10 @@ type TaskCanceled = ContractClient.Event<{
 type RewardPayoutCycleStarted = ContractClient.Event<{
   payoutId: number, // The payout ID logged when a new reward payout cycle has started.
 }>;
+type ColonyLabelRegistered = ContractClient.Event<{
```
coyotespike: This event is also in `ColonyNetwork`, below. Should it be in both files?

packages/colony-js-client/src/ColonyNetworkClient/index.js
JamesLefrere: Good find!

packages/colony-js-client/src/ColonyClient/index.js
JamesLefrere: In this case, yes, because it's emitted on the network client, but the method is called from the colony client (but it can be subscribed on both).

PR #285, opened 2018-10-18
Supply options for `createToken` (make it a `Sender`)

JamesLefrere:
## Description

This PR makes the `createToken` method work as a regular `Sender`.

Before:

```js
const address = await networkClient.createToken({ name, symbol });
```

After:

```js
const { meta: { receipt: { contractAddress } } } = await networkClient.createToken.send(
  {
    name,
    symbol,
  },
  options,
);
```

This means that we can supply options for the transaction that we send and also estimate it. Hopefully it makes things a little more predictable (at the expense of a more complicated first step).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## v.NEXT
 
+**Breaking changes**
+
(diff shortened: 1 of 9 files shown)


sprusr:
@JamesLefrere this is a nice move! All looks good to me too (but already merged, so 😎)

PR #289, opened 2018-10-20
Handle events not on interface (incl. `Transfer`, `Mint`)

### LINKED ISSUE #288: Cannot read property 'topics' of undefined
When attempting to use `colonyClient.mintTokens()` I am receiving the following error:

```
TypeError: Cannot read property 'topics' of undefined
```

The error is being thrown from:

```
/node_modules/@colony/colony-js-contract-client/lib/classes/ContractClient.js:273:46
```

In `ContractClient`, when `getReceiptEventData` is called `events[name]` is `undefined`. I was able to successfully use `mintTokens` when returning `null` if `events[name]` is `undefined`.

JamesLefrere:
## Description

This PR adds support for some events that are emitted from transactions on `ColonyClient` but originate from another contract (`Token.sol`). The approach taken is to simply copy over the defined event interface from the `token` (`TokenClient`) property of the `ColonyClient` to itself, as was done previously for events originating on the network contract. Though not ideal, I think this is a more reasonable approach than changing `IColony.sol` (I'm not convinced that a file like this should have a copy of all event definitions from contracts it calls methods on).

## Other changes

* Handles an edge case where an event exists in the logs but is not available on the contract interface; this could happen with e.g. BYOT (Bring Your Own Token), where additional events are defined. The missing events are ignored (but are still available in a raw format in the receipt logs).

Resolves #288

### DIFF
--- a/docs/_API_ColonyClient.md
+++ b/docs/_API_ColonyClient.md
@@ -253,7 +253,7 @@ A promise which resolves to an object containing the following properties:
 |Return value|Type|Description|
 |---|---|---|
 |address|Address|Address of the user for the given role.|
-|rated|boolean|Has the user work been rated.|
+|rateFail|boolean|Whether the user failed to rate their counterpart.|
 |rating|number|Rating the user received (1-3).|
 
 ### `getTaskWorkRatings.call({ taskId })`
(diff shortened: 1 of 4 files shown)


docs/_API_ColonyClient.md
```diff
@@ -253,7 +253,7 @@ A promise which resolves to an object containing the following properties:
 |Return value|Type|Description|
 |---|---|---|
 |address|Address|Address of the user for the given role.|
-|rated|boolean|Has the user work been rated.|
+|rateFail|boolean|Whether the user failed to rate their counterpart.|
```
JamesLefrere: Surprised this stuff is here; looks like maybe a rebase on the previous PR didn't have the changes from git hooks (docs update).

ryanchristo: Well, at least they're in there now. :+1:

JamesLefrere:
cc @ryanchristo Realised I missed some important things; the actual `ContractEvent` instances (i.e. `colonyClient.events.Transfer` and `colonyClient.events.Mint`), and the `from` property for `Transfer` from the previous issue.

PR #290, opened 2018-10-20
Add token events conditionally

JamesLefrere:
## Description

Fixes an issue where events from `token` were added to the `ColonyClient` when the token wasn't set.


JamesLefrere:
Tests pass, that'll do

PR #292, opened 2018-11-07
Fix `ColonyNetworkClient.lookupRegisteredENSDomain` call

JamesLefrere:
## Description

Fix the function called by `ColonyNetworkClient.lookupRegisteredENSDomain`.


JamesLefrere:
See: https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/IColonyNetwork.sol#L252

PR #299, opened 2018-11-21
Fix expected Buffer type from `getTask` result

ryanchristo:
## Description

When converting output values for `getTask`, the expected result is a `Buffer`. This pull request coverts the result to the expected `Buffer` type before returning the task.

## Dependencies

**New dependencies**:

```
"buffer": "^5.1.0"
```

## Resolves

JoinColony/colonyStarter#43


ryanchristo:
This fix applies to colonyNetwork `d50abbe` (latest Rinkeby deployment) and `f56cb3e` (latest).

PR #306, opened 2018-11-29
Update token name and symbol param type

### LINKED ISSUE #305: getTokenInfo throws "Invalid UTF-8 detected" error
## Description

When attempting to use `getTokenInfo`, I am receiving the following error:

```
Error: Invalid UTF-8 detected
```

ryanchristo:
## Description

This pull request updates the token `name` and `symbol` param type to `hexString`, which resolves the `Invalid UTF-8 detected` error when attempting to use the `getTokenInfo` method.

## TODO

- [ ] Maybe convert the `hexString`

## Resolves

#305


JamesLefrere:
Thanks Ryan, but I think a better approach is to handle this with #310 , where invalid UTF8 hex strings (that could/should be valid) are returned as-is.

ryanchristo:
Closing in favor of #310

PR #307, opened 2018-11-29
Sunset `setUserRole` and `getUserRoles`

### LINKED ISSUE #304: Authority client methods need to be updated
## Description

`colonyClient.authority.setUserRole` is always unsuccessful. This is either outdated and needs to be removed from the docs or a bug that needs to be fixed.

I am able to successfully manage user roles with the following methods `colonyClient.setOwnerRole`, `colonyClient.setAdminRole`, and `colonyClient.removeAdminRole`.

JamesLefrere:
## Description

👋 

(This method is not supported by the contracts anymore).

Resolves #304 and #200

### DIFF
--- a/docs/_API_ColonyClient.md
+++ b/docs/_API_ColonyClient.md
@@ -470,7 +470,7 @@ An instance of a `ContractResponse`
 
 ### `setRewardInverse.send({ rewardInverse }, options)`
 
-Set new colony recovery role. Can only be called by the founder role.
+Set the reward inverse to pay out from revenue. e.g. if the fee is 1% (or 0.01), set 100
 
 **Arguments**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
   * `removeRecoveryRole`
   * `setRecoveryRole`
   * `setStorageSlotRecovery`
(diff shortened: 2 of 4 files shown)


docs/_API_ColonyClient.md
```diff
@@ -470,7 +470,7 @@ An instance of a `ContractResponse`
 
 ### `setRewardInverse.send({ rewardInverse }, options)`
 
-Set new colony recovery role. Can only be called by the founder role.
+Set the reward inverse to pay out from revenue. e.g. if the fee is 1% (or 0.01), set 100
```
JamesLefrere: Seems like docgen wasn't called before 🤔

ryanchristo: At least it's fixed now 👍

JamesLefrere:
We can do that... perhaps we should talk to the network team about exposing `hasUserRole`, so we can do away with this class.

JamesLefrere:
Let's see how https://github.com/JoinColony/colonyNetwork/pull/450 does.

JamesLefrere:
I guess we can now ditch `AuthorityClient` altogether?

JamesLefrere:
Closing in favour if #311

PR #311, opened 2018-11-30
Remove `AuthorityClient`

JamesLefrere:
## Description

* Remove `AuthorityClient` as it is no longer needed
* Add `hasUserRole` caller to `ColonyClient`

## Other changes

* Rename `OWNER` role to `FOUNDER`


packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -248,6 +248,19 @@ export default class ColonyClient extends ContractClient {
...
+  */
+  hasUserRole: ColonyClient.Caller<
+    {
+      user: Address, // The user in question.
+    },
+    {
+      role: AuthorityRole, // That user's role (`FOUNDER` or `ADMIN`).
+    },
+    ColonyClient,
+  >;
```
ryanchristo: ☝️ should be...

```
  hasUserRole: ColonyClient.Caller<
    {
      user: Address, // The user in question.
      role: AuthorityRole, // That user's role (`FOUNDER` or `ADMIN`).
    },
    {
      hasRole: boolean, // Whether the user has the given role
    },
    ColonyClient,
  >;
```

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -541,7 +554,7 @@ export default class ColonyClient extends ContractClient {
     ColonyClient,
   >;
   /*
-    Set a new colony admin role. Can be called by an owner or admin role.
+    Set a new colony admin role. Can be called by an founder or admin role.
```
ryanchristo: ☝️ should be "a founder"

JamesLefrere: an good find.

PR #320, opened 2018-12-13
Add event log parsing

JamesLefrere:
## Description

This PR adds historical event log parsing functionality:

* With a `ContractEvent` instance, given an array of logs, filter matching topics and parse event data from them
* Add a `ContractEvent` getter for the underlying event interface
* Add the `getLogs` method to the `Provider` interface
* Fix the event interface `parse` type

## TODO

- [x] Consider adding a `parseLogs` method on `ContractClient`
- [x] Consider adding a `getLogs` method on `ContractClient` (maybe also parsing the logs?)

### DIFF
--- a/packages/colony-js-contract-client/src/classes/ContractClient.js
+++ b/packages/colony-js-contract-client/src/classes/ContractClient.js
@@ -4,6 +4,8 @@ import type { Query } from '@colony/colony-js-contract-loader';
 import type {
   IAdapter,
   IContract,
+  Log,
+  LogFilter,
   TransactionOptions,
   TransactionReceipt,
 } from '@colony/colony-js-adapter';
@@ -32,6 +34,9 @@ export default class ContractClient {
 
   // The contract event subscription methods
   events = {};
+
+  // Mapping of event topics to ContractEvents
+  eventSignatures = {};
 
   // Static getters used in lieu of named exports; this package only has
   // one export.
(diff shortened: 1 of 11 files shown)


packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -194,12 +234,15 @@ export default class ContractClient {
...
+    });
+
     Object.assign(this.events, {
-      [eventName]: new ContractEvent({
-        eventName,
-        client: this,
-        argsDef,
-      }),
+      [eventName]: event,
+      [event.interface.topics[0]]: event,
```
JamesLefrere: I think this might be a little cleaner with a separate object, like `eventSignatures` 🤔 Just to keep this object clean/easily understandable.

packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -117,6 +119,35 @@ export default class ContractClient {
...
+  ): Promise<Array<Object>> {
+    const { eventNames = [], topics = [] } = filter;
+
+    // Get topics for the given eventNames
+    const extraTopics: Array<string> = eventNames
+      .filter(eventName => Object.keys(this.events).includes(eventName))
+      .map(eventName => this.events[eventName].interface.topics[0]);
+
+    // Combine any existing topics with the extra ones
+    if (!topics.length) {
```
JamesLefrere: This seems complicated; can we flatten unique items like this?

```
const allTopics = [...new Set([...topics, ...extraTopics])];
```

packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -117,6 +119,35 @@ export default class ContractClient {
...
+   * Get logs from the contract with filter, return parsed event logs.
+   */
+  async getLogs(
+    filter?: LogFilter & { eventNames?: Array<string> } = {},
+  ): Promise<Array<Object>> {
+    const { eventNames = [], topics = [] } = filter;
+
+    // Get topics for the given eventNames
+    const extraTopics: Array<string> = eventNames
+      .filter(eventName => Object.keys(this.events).includes(eventName))
```
JamesLefrere: Maybe instead throw an error for events that don't exist? Otherwise it's not clear that I'm not getting what I asked for.

packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -117,6 +119,35 @@ export default class ContractClient {
...
+
+    // Combine any existing topics with the extra ones
+    if (!topics.length) {
+      topics.push(extraTopics);
+    } else if (Array.isArray(topics[0])) {
+      topics[0].push(...extraTopics);
+    } else {
+      topics[0] = [topics[0], ...extraTopics];
+    }
+    const augmentedFilter = {
```
JamesLefrere: Seems like an unnecessary variable :)

packages/colony-js-contract-client/src/classes/ContractClient.js
sprusr: Yeah alright 👍 do you reckon a mapping from `topic` to `ContractEvent` or just `eventName`? I'm very indifferent 🤷‍♂️

JamesLefrere: The former sounds good 👍

JamesLefrere:
LGTM!

Issue #248, opened 2018-08-26
Add specific colonyJS package versions and update colonyNetwork commit


igorline:
I was trying to run the example described in "Get Started" but the process froze after creating a token and a colony. The problem was fixed after setting `colony-js-adapters-ethers` version to `1.5.3` and reinstalling the dependencies. This should be reflected in the documentation. - @igorline
_____

The original issue has been solved but I'm going to propose we include specific versions for installing the colonyJS packages to prevent this issue in the future. Also, going to note here that the commit for the colonyNetwork needs to be updated. - @ryanchristo 

- [ ] Update colonyNetwork commit
- [ ] Add specific versions to colonyJS packages

ryanchristo:
@igorline Thanks for posting and sorry for the inconvenience! There is a pending pull request (#244) to add specific versions to the example. I will ping the request and we'll get that merged here soon.

ryanchristo:
The original issue has been resolved with update colonyJS packages but I'm going to keep this open as a reminder to set specific versions in the "Getting Started" section of the docs to prevent this issue in the future. Added note to description.

PR #331, opened 2018-12-20
Update versions in "Get Started" doc

### LINKED ISSUE #344: Error: No such event "UserLabelRegistered" on this contract
Not sure where's the best place to report this. I'm following the [Get Started](https://docs.colony.io/colonyjs/docs-get-started/) guide and coming across the error:

```
Error: No such event "UserLabelRegistered" on this contract
```

when running the sample code.

Here's are the steps I took:

```bash
docker pull ethereum/solc:0.4.23

git clone --recursive https://github.com/JoinColony/colonyNetwork.git
cd colonyNetwork

git checkout d50abbeb9f119850cb70e9ec854576123a707205

yarn

yarn global add trufflepig

# in separate terminal
./node_modules/.bin/ganache-cli -d --gasLimit 7000000 --acctKeys ganache-accounts.json

./node_modules/.bin/truffle migrate --reset --compile-all

trufflepig --ganacheKeyFile ganache-accounts.json
```

```bash
mkdir colonyExample

cd colonyExample

yarn init

yarn add @colony/colony-js-adapter-ethers@1.7.0 @colony/colony-js-client@1.7.5 @colony/colony-js-contract-loader-http@1.6.2 ethers
```

`index.js`

```js
// Import the prerequisites
const { providers, Wallet } = require('ethers');
const { default: EthersAdapter } = require('@colony/colony-js-adapter-ethers');
const { TrufflepigLoader } = require('@colony/colony-js-contract-loader-http');
const { default: ColonyNetworkClient } = require('@colony/colony-js-client');

// Create an instance of TrufflepigLoader
const loader = new TrufflepigLoader();

// Create an instance of JsonRPCProvider using the url of our test network
const provider = new providers.JsonRpcProvider('http://localhost:8545/');

// Create an async function
const example = async () => {

  // Get the private key from the first account
  const { privateKey } = await loader.getAccount(0);

  // Create an instance of Wallet using the private key and provider
  const wallet = new Wallet(privateKey, provider);


  // Create an instance of EthersAdapter
  const adapter = new EthersAdapter({
    loader,
    provider,
    wallet,
  });

  // Create an instance of ColonyNetworkClient using the adapter
  const networkClient = new ColonyNetworkClient({ adapter });

  // Initialize the client
  await networkClient.init();

  // Congrats, you've connected to the network!
  console.log('network address: ' + networkClient.contract.address);

  // Create an ERC20 token (you could also skip this step and use a pre-existing token)
  const { meta: { receipt: { contractAddress } } } = await networkClient.createToken.send({
    name: 'Token',
    symbol: 'TKN',
  });

  // Congrats, you've created an ERC20 token!
  console.log('token address: ' + contractAddress);

  // Create a colony using the token address of the ERC20 token we created
  const { eventData: { colonyId, colonyAddress } } = await networkClient.createColony.send({
    tokenAddress: contractAddress,
  });

  // Congrats, you've created a colony!
  console.log('colony id: ' + colonyId);
  console.log('colony address: ' + colonyAddress);

  // Get an initialized ColonyClient for the colony we just created
  const colonyClient = await networkClient.getColonyClient(colonyId);

};

// Execute the example async function
example()
  .then(() => process.exit())
  .catch(error => console.error(error));
```

```bash
node index.js
```

Any help or guidance is appreciated

ryanchristo:
## Description

This pull request updates the colonyNetwork commit to `f73dc84a41f5fc1962c999a24e13b15ba491b8a6` in the "Get Started" page of the docs and removes the specific colonyJS package versions.

_NOTE: The specific versions for colonyJS packages were removed because the latest versions of the packages will need to support all versions of the contracts in the future, therefore the latest colonyJS packages should always be recommended when getting started. (tagging #248)_

_NOTE: There is a known issue using this version of the colonyNetwork contracts with the current `rinkeby` deployment but "Get Started" is focused on setting up a local development environment and the current version of the docs reflect the latest colonyJS package versions._

## Resolves

#344

### DIFF
--- a/docs/_Docs_GetStarted.md
+++ b/docs/_Docs_GetStarted.md
@@ -36,7 +36,7 @@ docker pull ethereum/solc:0.4.23
 
 ### colonyNetwork
 
-Using colonyJS in a local development environment will require a locally deployed version of the colonyNetwork smart contracts, so the first order of business will be pulling down the colonyNetwork repository.
+Using colonyJS in a local development environment will require a locally deployed version of the colonyNetwork smart contracts.
 
 In the working directory of your choice, clone the latest version of [colonyNetwork](https://github.com/JoinColony/colonyNetwork):
 
@@ -49,9 +49,20 @@ git clone --recursive https://github.com/JoinColony/colonyNetwork.git
 ```
 cd colonyNetwork
 
-git checkout d50abbeb9f119850cb70e9ec854576123a707205
+git checkout f73dc84a41f5fc1962c999a24e13b15ba491b8a6
 
 yarn
+
+```
+
+Next we'll need to initialize and provision the colonyNetwork submodules:
+
+```
+
+git submodule update --init --recursive
+
+yarn provision:token:contracts
+
 ```
 
 ### TrufflePig
@@ -71,7 +82,7 @@ For our local test network, there are a few tweaks to default settings that need
 With `ganache-cli` we can do this all in one command at start. Open up a new terminal window and, within the colonyNetwork directory, run the following command:
 
 ```
-./node_modules/.bin/ganache-cli -d --gasLimit 7000000 --acctKeys ganache-accounts.json
+yarn start:blockchain:client
 ```
 
 This will start up a new test network that keeps account keys in a place that is easy to access for TrufflePig.
@@ -113,10 +124,8 @@ yarn init
 Add the required libraries to your project with `yarn`:
 
 ```
-yarn add @colony/colony-js-adapter-ethers@1.7.0 @colony/colony-js-client@1.7.5 @colony/colony-js-contract-loader-http@1.6.2 ethers
+yarn add @colony/colony-js-adapter-ethers @colony/colony-js-client @colony/colony-js-contract-loader-http ethers
 ```
-
-*As you can see in the above command, we have included specific colonyJS package versions. These are recommended versions that have been tested with the colonyNetwork version that we deployed to our local test network.*
 
 ## Create a new colony
 
--- a/docs/_Docs_GetStarted.md
+++ b/docs/_Docs_GetStarted.md
@@ -77,12 +77,24 @@ yarn global add trufflepig
 
 ## Fire up a test network
 
-For our local test network, there are a few tweaks to default settings that need to be made: We want to set the `gasLimit` to `7000000` and we want all our account keys to be stored in a `.json` file that we'll be able to easily call inside our application.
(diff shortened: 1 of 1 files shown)


docs/_Docs_GetStarted.md
```diff
@@ -110,13 +133,17 @@ cd colonyExample
...
-Add the required libraries to your project with `yarn`:
+Add the required colonyJS libraries to your project with `yarn`:
 
 ```
-yarn add @colony/colony-js-adapter-ethers@1.7.0 @colony/colony-js-client@1.7.5 @colony/colony-js-contract-loader-http@1.6.2 ethers
+yarn add @colony/colony-js-adapter-ethers @colony/colony-js-client @colony/colony-js-contract-loader-http
 ```
 
-*As you can see in the above command, we have included specific colonyJS package versions. These are recommended versions that have been tested with the colonyNetwork version that we deployed to our local test network.*
+Next, we will need to require a specific version of the `ethers` package to work with the version of `ganache-cli` that we have installed in the version of colonyNetwork that we are using in this example.
```
sprusr: Just a clarification, I think the specific Ethers version is because colonyJS’s EthersAdapter doesn’t support v4, rather than it being as a result of ganache. I think.

docs/_Docs_GetStarted.md
```diff
@@ -110,13 +133,17 @@ cd colonyExample
 yarn init
 ```
 
-Add the required libraries to your project with `yarn`:
+Add the required colonyJS libraries to your project with `yarn`:
 
 ```
-yarn add @colony/colony-js-adapter-ethers@1.7.0 @colony/colony-js-client@1.7.5 @colony/colony-js-contract-loader-http@1.6.2 ethers
+yarn add @colony/colony-js-adapter-ethers @colony/colony-js-client @colony/colony-js-contract-loader-http
```
sprusr: Is it okay to just use the latest version? When colonyJS is upgraded, if we don’t also update the colonyNetwork hash in here then there is potentially an API mismatch. But maybe it is okay?

docs/_Docs_GetStarted.md
ryanchristo: Hmmmm.. Yea, it might be the other way around. I came across an old issue that said it was a mismatch between `ethers` and `ganache-cli` but I think you're right.

docs/_Docs_GetStarted.md
ryanchristo: Yea, this is an ongoing question and depends on what we decide for #340. I think it would be nice to not have version tags, which would require us to update this doc each time we upgrade colonyJS to ensure people are using the latest colonyJS packages but that also runs the risk of a new colonyJS version not cooperating with the version of colonyNetwork specified here.

ryanchristo: I think this is ok for now and we'll need to circle back soon enough.

PR #332, opened 2018-12-21
Update and add contract client methods

### LINKED ISSUE #313: Implement `bootstrapColony` method
## Description

Implement `bootstrapColony` method.

https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/IColony.sol#L92

ryanchristo:
## Description

This pull request adds the `bootstrapColony` method and the `ColonyBootstrapped` event. In order to add the bootstrap method and event, I had to also add the `array` type to `paramTypes`.

## Other changes

This pull request also fixes the output type for the `TaskRoleUserSet` event.

## Resolves

#313

### DIFF
--- a/packages/colony-js-contract-client/src/modules/paramTypes.js
+++ b/packages/colony-js-contract-client/src/modules/paramTypes.js
@@ -10,7 +10,7 @@ import {
 } from '@colony/colony-js-utils';
 
 import type { ParamTypes, ParamTypeDef } from '../flowtypes';
-import { isBoolean } from './paramValidation';
+import { isArray, isBoolean } from './paramValidation';
 
 const passThrough = value => value;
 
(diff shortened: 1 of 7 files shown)


packages/colony-js-contract-client/src/modules/paramTypes.js
```diff
@@ -29,6 +29,13 @@ const PARAM_TYPE_MAP: {
     },
     convertInput: passThrough,
   },
+  array: {
```
sprusr: I'm thinking this probably wants to be more specific. I did a bit of stuff with arrays in vesting client, but not sure how applicable that is here: https://github.com/JoinColony/vestingClient/blob/master/src/paramTypes.js#L8

ryanchristo: Thanks! I was not entirely sure what to do here. I'll take a look at the vestingClient.

PR #334, opened 2018-12-29
Fix links to multisig docs

igorline:
Currently in colonyjs docs all the links to multisig description are leading to non-existing page, this PR should fix the issue.


ryanchristo:
Thanks @igorline!

Issue #340, opened 2019-01-09
Versioning for Colony contract


chmanie:
### Description
Apart from versioning the general contracts, we want to enable the colony founders to upgrade their colony to the next version. This is totally in their power so we have to find a way to be backward compatible to the old versions of the colony contract.

### Thoughts
This cuts very deep into the ColonyJS stack. It probably even goes as far as the loader level. The problem here is that the interface of IColony.json is also reflected statically in ColonyJS. Does that mean we have to also version the ColonyClient itself and copy multiple versions of it? This seems to be the easiest solution for now but maybe someone else has a better idea. This will probably be somewhat easier to do with tailor.

Keep in mind that we can use [`getVersion`](
https://docs.colony.io/colonyjs/api-colonyclient/#getversioncall) to get the currently deployed version of that particular colony.

The goal here is to have something that is ~working and can be replaced later with a more sophisticated implementation when ColonyJS 2.0 is out (based on tailor).

### Notes
This also means, we have to keep the old compiled versions (.json) of the contracts around, so this will influence the contract deployment part as well. Is there something planned for that @area, @elenadimitrova? Maybe the assignee of this issue and you could coordinate.

chmanie:
DONE!

PR #346, opened 2019-01-23
Add contract info and update client methods

ryanchristo:
## Description

This pull request adds contract information to the client methods and events, which is used to display information in the docs about what contract the colonyJS methods and events are associated with.

In the process of looking up the methods and events on the contract, I found some things that need to be updated/removed, which I've added to the checklist below and will address here.

Thinking in terms of colonyJS supporting multiple colony contract versions, the contract information could contain colony contract versions that would be used to filter out methods and events that need to be loaded into the colony client, therefore different colonies using different colony contract versions would only get the methods and events that related to the version they are using. This is not included in this pull request but I figured it would be worth mentioning here. (#340)

#### ColonyClient

- [X] Address `assignWorkRating` (internal function)
  * Removed in commit `c4bc4a9`
- [X] Address `getGlobalRewardPayoutCount` (no matching contract methods)
  * Updated to `getTotalLockCount` in commit `58efcf8`
- [X] Address `getRecoveryRolesCount` (no colonyJS method, no matching contract methods)
  * Updated contract data in commit `ff91460`
- [X] Address `getTransactionCount` (no colonyJS method, no contract interface)
  * Removed in commit `e392f72`
- [X] Address `getUserRewardPayoutCount` (no matching contract methods)
  * Update to `getUserLock` in commit `0efb6b1`
- [X] Address `setToken` (no matching contract methods)
  * Removed in commit `8e8abef`
- [X] Address `waiveRewardPayouts` (no matching contract methods)
  * Updated to `incrementLockCounterTo` in commit `c1f4b58`
- [X] Address `initialise` (no colonyJS spec, should this be here?)
  * Removed in commit `5818b55`

#### ColonyNetworkClient

- [X] Address `getRecoveryRolesCount` (no colonyJS method, no matching contract methods)
  * Updated contract data in commit `ff91460`

#### MetaColonyClient

- [X] Address `getAuthority` (should this be here? already included in colony client)
  * Removed in commit `813b3b4`
- [ ] Address `getToken` (should this be here? already included in colony client)
  * ___This needs to be addressed in another PR.___
- [X] Address `setNetworkFeeInverse` (no colonyJS method)
  * Added in commit `8200095`

### DIFF
--- a/packages/colony-js-client/src/MetaColonyClient/index.js
+++ b/packages/colony-js-client/src/MetaColonyClient/index.js
@@ -39,8 +39,15 @@ export default class MetaColonyClient extends ContractClient {
     {
       parentSkillId: number, // The numeric ID of the skill under which the new skill will be added.
     },
-    { SkillAdded: SkillAdded },
+    {
+      SkillAdded: SkillAdded,
+    },
     MetaColonyClient,
+    {
+      contract: 'Colony.sol',
+      interface: 'IMetaColony.sol',
+      version: 0,
+    },
   >;
   /*
   Get the authority contract address associated with the colony.
@@ -51,6 +58,12 @@ export default class MetaColonyClient extends ContractClient {
       address: Address, // The address of the authority contract associated with the colony.
     },
     MetaColonyClient,
+    {
+      name: 'authority',
+      contract: 'dappsys/auth.sol',
+      interface: 'IColony.sol',
+      version: 0,
+    },
   >;
   /*
   Get the address of the ERC20 token contract that is the native token assigned to the Meta Colony.
@@ -61,6 +74,11 @@ export default class MetaColonyClient extends ContractClient {
       address: Address, // The address of the ERC20 token contract.
     },
     MetaColonyClient,
+    {
+      contract: 'Colony.sol',
+      interface: 'IColony.sol',
+      version: 0,
+    },
   >;
   /*
   Mint tokens for the Colony Network. This can only be called from the Meta Colony, and only by the user assigned the `FOUNDER` role.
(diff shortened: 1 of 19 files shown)


ryanchristo:
screen shot 2019-01-24 at 12 05 00 pm

packages/colony-js-client/src/MetaColonyClient/index.js
```diff
@@ -39,18 +39,15 @@ export default class MetaColonyClient extends ContractClient {
...
-  Get the authority contract address associated with the colony.
-  */
-  getAuthority: MetaColonyClient.Caller<
-    {},
     {
-      address: Address, // The address of the authority contract associated with the colony.
+      SkillAdded: SkillAdded,
     },
     MetaColonyClient,
+    {
```
JamesLefrere: This is a bit of a perversion of generics, but we're already using these and the comments above for documentation, so I guess I can live with it; it's fairly easy to manage.

packages/colony-js-client/src/TokenClient/index.js
```diff
@@ -98,6 +128,12 @@ export default class TokenClient extends ContractClient {
       decimals: number, // The number of decimals.
     },
     TokenClient,
+    {
+      contract: 'name, symbol, decimals',
```
JamesLefrere: Is this right? 🤔

packages/colony-js-client/src/addTokenLockingMethods.js
```diff
@@ -0,0 +1,24 @@
+/* @flow */
+
+import ContractClient from '@colony/colony-js-contract-client';
+
+/*
+ * Add methods from `ITokenLocking.sol` to a given `ContractClient`.
+ */
+const addTokenLockingMethods = (client: ContractClient) => {
```
JamesLefrere: Where is this called?

ryanchristo: [ColonyClient/index.js](https://github.com/JoinColony/colonyJS/blob/2d672ac083d8972222975ac6bb84918800d2c130/packages/colony-js-client/src/ColonyClient/index.js#L1347)

packages/colony-js-client/src/MetaColonyClient/index.js
ryanchristo: I hope it's not too much trouble. I think it will do more good than harm. 😅

packages/colony-js-client/src/addTokenLockingMethods.js
JamesLefrere: Ok, nice 👍

PR #347, opened 2019-01-24
Various updates based on contract changes

### LINKED ISSUE #338: Changes to Token contract
In the upcoming https://github.com/JoinColony/colonyNetwork/pull/479, the `ERC20ExtendedToken` is removed in favour of the Dappsys `DSToken`.

The `ERC20ExtendedToken` constructor accepted `name`, `symbol` and `decimals`, whereas `DSToken` accepts only `symbol`, with hardcoded `decimals = 18` and name optionally being later set with `setName`.

Both `symbol` and `name` were previously of type `string`, but are now `bytes32`. We will likely need to handle the conversion to/from hex in this library (before/after ethers).

- [ ] New Token constructor
- [ ] Set Token name with `setName`
- [ ] Handle conversion of strings to/from hex strings (`bytes32`) for `name` and `symbol`

sprusr:
## Description

`TokenClient` is now based on the `DSToken` contract, due to the removal of `ERC20ExtendedToken` contract. The `ColonyTokenSet` event has also been removed.

Up to date as of colonyNetwork#9ddf14118cd0436a19a7278282e4b8a607e14b54.

Closes #338


packages/colony-js-client/src/ColonyNetworkClient/index.js
```diff
@@ -144,9 +144,7 @@ export default class ColonyNetworkClient extends ContractClient {
   */
   createToken: ColonyNetworkClient.Sender<
```
ryanchristo: Is it possible to make `name` and `decimals` optional or has this ability been completely removed with the new token contract?

ryanchristo:
I've been updating colonyJS (in #346 especially) to work with the following colonyNetwork commit:

`f73dc84a41f5fc1962c999a24e13b15ba491b8a6`

Should we be holding off on updates to later commits or should I be updating to the latest commit?

packages/colony-js-client/src/ColonyNetworkClient/index.js
sprusr: `decimals` is now hardcoded, and `name` is set with `setName`. It is quite annoying! But also a more widely used contract, so perhaps more expected.

packages/colony-js-contract-client/src/modules/paramTypes.js
```diff
@@ -70,6 +76,18 @@ const PARAM_TYPE_MAP: {
...
+  bytes32String: {
+    validate(value: any) {
+      return typeof value === 'string';
+    },
+    convertOutput(value) {
+      return hexToUtf8(value);
+    },
+    convertInput(value) {
+      const converted = utf8ToHex(value);
+      return converted.padEnd(66, '0');
```
chmanie: Can you explain what’s going on here? I guess a comment wouldn’t hurt, either.

chmanie:
@sprusr you said that there’s not going to be any changes around the dApp for the token creation?

sprusr:
@chmanie no there is, we will need to add an extra transaction to do `setName`. But otherwise it’s the same.

packages/colony-js-contract-client/src/modules/paramTypes.js
sprusr: Ah you actually made me realise this is totally wrong... So for `bytes32` values, the underlying ethers adapter wants a full 32-byte hex string (64 characters + `0x`). What I've done here is pad it out at the wrong end though. It should pad with zeros from the start.

Will fix and add comment.

PR #351, opened 2019-01-30
Fix and update colony client methods and events

ryanchristo:
## Description

This pull request includes minor fixes to colony client events, the addition of missing colony client methods, and fixes and adds missing types in `docgen`.

Also, this pull request fixes the structure of the clients in relation to adding methods from interfaces not set as the default interface in the client, which was a mistakes I made in #346 and #350.

As a result of the fixes, `TokenLockClient` has been added and documentation for `TokenLockClient` will be updated in #348.


packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1331,26 +1327,10 @@ export default class ColonyClient extends ContractClient {
...
-      contract: 'TokenLocking.sol',
-      interface: 'ITokenLocking.sol',
-      version: 'f73dc84a41f5fc1962c999a24e13b15ba491b8a6',
-    },
-  >;
 
   static get defaultQuery() {
     return {
-      contractName: 'IColony',
+      contractName: 'IMetaColony', // IMetaColony extends IColony
```
JamesLefrere: I'd suggest leaving this one as-is, and in the network client, doing this:

```js
  async getMetaColonyClientByAddress(contractAddress: Address) {
    const metaColonyClient = new this.constructor.ColonyClient({
      adapter: this.adapter,
      networkClient: this,
      query: { contractAddress, contractName: 'IMetaColony' },
    });
    await metaColonyClient.init();
    return metaColonyClient;
  }
```

The contract name should override the default.

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1599,15 +1576,20 @@ export default class ColonyClient extends ContractClient {
     if (this.token) {
       this.events.Transfer = this.token.events.Transfer;
       this.events.Mint = this.token.events.Mint;
+      this.events.TokenLocked = this.token.events.TokenLocked;
```
JamesLefrere: As discussed, let's make sure we get these from the right place.

packages/colony-js-client/src/constants.js
```diff
@@ -4,7 +4,7 @@ export const WORKER_ROLE = 'WORKER';
 export const EVALUATOR_ROLE = 'EVALUATOR';
 export const MANAGER_ROLE = 'MANAGER';
 
-export const ROLES = {
+export const TASK_ROLES = {
```
JamesLefrere: 👍

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1599,15 +1576,20 @@ export default class ColonyClient extends ContractClient {
...
       this.contract.interface.events.Mint = this.token.contract.interface.events.Mint;
+      this.contract.interface.events.TokenLocked = this.token.contract.interface.events.TokenLocked;
     }
     /* eslint-enable max-len */
 
     // Senders
     this.addSender('addDomain', {
       input: [['parentDomainId', 'number']],
     });
+    this.addSender('addGlobalSkill', {
```
JamesLefrere: I'm not sure about these being available for non-metacolony client instances... how about something like this:

```
// add a getter to this class
get isMetaColony() {
  return this._query.contractName === 'IMetaColony';
}

// here, use it to add these conditionally:

if (this.isMetaColony) {
  this.addSender('addGlobalSkill', { /* ... */ });
  // ... other MetaColony methods
}
```

Maybe?

JamesLefrere: Or more forgiving:

```js
addMetaColonySender(...args: *) {
  return this.isMetaColony ? this.addSender(...args) : this._metaColonyOnlyError(...args);
}

_metaColonyOnlyError(propName: string) {
  this[propName] = () => { throw new Error(`"${propName}" is only available for the MetaColony`) };
}
```

Or something?

packages/colony-js-client/src/TokenLockingClient/index.js
```diff
@@ -0,0 +1,241 @@
...
+  token: TokenAddress, // The address of the token contract that was unlocked.
+  user: Address, // The address of the user that the tokens were unlocked for.
+  lockId: number, // The numeric ID of the lock that the was set for the user.
+}>;
+type UserTokenWithdrawn = ContractClient.Event<{
+  token: TokenAddress, // The address of the token contract from which tokens were withdrawn.
+  user: Address, // The address of the user that withdrew tokens.
+}>;
+
+export default class ColonyClient extends ContractClient {
```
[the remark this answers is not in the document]
ryanchristo: Whoops! Yup, will fix.

packages/colony-js-client/src/ColonyClient/index.js
ryanchristo: Aha! This is what I wanted to do. Much better.

packages/colony-js-client/src/ColonyClient/index.js
ryanchristo: Yup, working on it. 👍

packages/colony-js-client/src/constants.js
ryanchristo: Yea, that was confusing with authority roles, recovery roles, and task roles.

packages/colony-js-client/src/ColonyClient/index.js
ryanchristo: Good idea. I will explore the options you shared.

PR #352, opened 2019-02-06
Add `getNetworkClient` to colony-js-client

ryanchristo:
## Description

This is a bit of an experiment and a potential temporary solution to making the connection process a bit easier for developers. The `getNetworkClient` method proposed here is using `NetworkLoader` and `EthersAdapter`, additional inputs could be added to make this more dynamic but the target is to simplify the process for the "Get Started" page, which connects to the `rinkeby` test network.

The method supports the following use:
```
const { providers, Wallet } = require('ethers');
const { getNetworkClient } = require('@colony/colony-js-client');

const network = 'rinkeby';
const privateKey = '0x0...';

const connectNetwork = async () => {
  const provider = providers.getDefaultProvider(network);
  const wallet = new Wallet(privateKey, provider);
  const networkClient = await getNetworkClient(network, wallet);
  return networkClient;
};

```

Ideally the method would also support the following use (getting an error at the moment):
```
const { getNetworkClient } = require('@colony/colony-js-client');
const { open } = require('@colony/purser-software');

const network = 'rinkeby';
const privateKey = '0x0...';

const connectNetwork = async () => {
  const wallet = await open({ privateKey });
  const networkClient = await getNetworkClient(network, wallet);
  return networkClient;
};

```

## Dependencies

**Added to `colony-js-client`**:

```
"@colony/colony-js-adapter-ethers": "^1.10.0",
"@colony/colony-js-contract-loader-network": "^1.6.2",
"ethers": "3.0.27",
```

### DIFF
--- a/packages/colony-js-client/src/getNetworkClient.js
+++ b/packages/colony-js-client/src/getNetworkClient.js
@@ -8,6 +8,7 @@ import ColonyNetworkClient from './ColonyNetworkClient/index';
 const getNetworkClient = async (network: string, wallet: any) => {
   const loader = new NetworkLoader({ network });
   const provider = providers.getDefaultProvider(network);
+  if (!wallet.provider) Object.assign(wallet, { provider });
   const adapter = new EthersAdapter({
     loader,
     provider,
--- a/packages/colony-js-client/src/getNetworkClient.js
+++ b/packages/colony-js-client/src/getNetworkClient.js
@@ -1,0 +1,21 @@
+/* @flow */
+
+import { providers } from 'ethers';
+import EthersAdapter from '@colony/colony-js-adapter-ethers';
+import NetworkLoader from '@colony/colony-js-contract-loader-network';
+import ColonyNetworkClient from './ColonyNetworkClient/index';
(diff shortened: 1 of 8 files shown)


packages/colony-js-client/src/getNetworkClient.js
```diff
@@ -0,0 +1,25 @@
...
+
+const getNetworkClient = async (network: string, wallet: any) => {
+  const loader = new NetworkLoader({ network });
+  const provider = providers.getDefaultProvider(network);
+  const adapter = new EthersAdapter({
+    loader,
+    provider,
+    wallet,
+  });
+  const query = {
```
JamesLefrere: I think this shouldn't be necessary (these are the default values). If flow is complaining, try passing an empty object for `query`.

ryanchristo: Yup, flow was complaining. I will try an empty object.

ryanchristo:
Yes, documentation changes will be made. Just wanted to make sure there was no red flags.

PR #357, opened 2019-02-12
Update to latest contracts and add one tx method

ryanchristo:
## Description

This pull request updates colonyJS methods and events to colonyNetwork commit [9bba127](https://github.com/JoinColony/colonyNetwork/commit/9bba127b0286708d4f8919526a943b0e916cfd7c) and adds the temporary one transaction payment method in the [temp/eth-denver](https://github.com/JoinColony/colonyNetwork/commits/temp/eth-denver) branch (commit [8888fd4](https://github.com/JoinColony/colonyNetwork/commit/8888fd47814ca9d24eb750c2f1b7ed24576a1744)).

## Checklist

- [x] Update colonyJS methods and events to colonyNetwork commit `9bba127`
- [x] Add the one transaction payment method in the `temp/eth-denver` branch

### DIFF
--- a/packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
+++ b/packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
@@ -1,0 +1,42 @@
+/* @flow */
+
+import ContractClient from '@colony/colony-js-contract-client';
+
+import type BigNumber from 'bn.js';
+import type ColonyClient from '../index';
+
+type Address = string;
+type TokenAddress = string;
+
+type InputValues = {
+  worker: Address,
+  token: TokenAddress,
+  amount: BigNumber,
+  domainId: number,
+  skillId: number,
+};
+
+type OutputValues = {
+  id: number,
+};
+
+export default class MakePayment extends ContractClient.Sender<
+  InputValues,
+  OutputValues,
+  ColonyClient,
+  *,
+> {
+  async _sendTransaction(args: *, options: *) {
+    const colony = this.client.contract.address;
+    const contract = await this._getContract(args);
+    args.unshift(colony);
+    return contract.callTransaction('makePayment', args, options);
+  }
+
+  async _getContract(args: *) {
+    return this.client.adapter.getContract(
+      { contractName: 'OneTxPayment' },
+      args,
+    );
+  }
+}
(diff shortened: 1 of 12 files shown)


packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
```diff
@@ -0,0 +1,42 @@
...
+
+export default class MakePayment extends ContractClient.Sender<
+  InputValues,
+  OutputValues,
+  ColonyClient,
+  *,
+> {
+  async _sendTransaction(args: *, options: *) {
+    const colony = this.client.contract.address;
+    const contract = await this._getContract(args);
```
JamesLefrere: I think this should be called without an argument.

packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
```diff
@@ -0,0 +1,42 @@
...
+    args.unshift(colony);
```
JamesLefrere: It's usually not a good idea to mutate function arguments, so I would suggest this change:

```js
    const contract = await this._getContract();
    return contract.callTransaction(
      'makePayment',
      // the first argument must be the colony address
      [this.client.contract.address, ...args],
      options,
    );
```

packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
```diff
@@ -0,0 +1,42 @@
...
+    return contract.callTransaction('makePayment', args, options);
+  }
+
+  async _getContract(args: *) {
+    return this.client.adapter.getContract(
+      { contractName: 'OneTxPayment' },
+      args,
```
JamesLefrere: Flow tells me this only accepts one argument (the query object), so I think what we need here is just the `contractName` – AFAIK there's only ever one `OneTxPayment` contract.

packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
ryanchristo: 👍

packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
ryanchristo: 👍

packages/colony-js-client/src/ColonyClient/senders/MakePayment.js
ryanchristo: 👍

PR #361, opened 2019-02-15
Add EthersWrappedWallet to colony-js-client

ryanchristo:
## Description

This pull request uses `EthersWrappedWallet` when passing a wallet instance to the adapter in the `getNetworkClient` method, allowing purser wallet instances to be used with `getNetworkClient`.

## Dependencies

**Added to `colony-js-client`**:

```
"@colony/purser-core": "^2.1.0",
"ethereumjs-tx": "^1.3.7",
```

### DIFF
--- a/packages/colony-js-client/src/getNetworkClient.js
+++ b/packages/colony-js-client/src/getNetworkClient.js
@@ -5,6 +5,7 @@ import EthersAdapter from '@colony/colony-js-adapter-ethers';
 import NetworkLoader from '@colony/colony-js-contract-loader-network';
 import { TrufflepigLoader } from '@colony/colony-js-contract-loader-http';
 import ColonyNetworkClient from './ColonyNetworkClient/index';
+import EthersWrappedWallet from './lib/EthersWrappedWallet/index';
 
 // This method provides a simple way of getting an initialized network client
 // that uses NetworkLoader for the remote network and TrufflePigLoader for a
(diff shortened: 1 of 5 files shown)


packages/colony-js-client/src/getNetworkClient.js
```diff
@@ -28,14 +29,12 @@ const getNetworkClient = async (network: string, wallet: any) => {
...
 
-  // Support offline wallets (wallets opened with purser)
-  if (!wallet.provider) Object.assign(wallet, { provider });
-
   // Initialize adpaters using ethers
   const adapter = new EthersAdapter({
     loader,
     provider,
-    wallet,
+    // $FlowFixMe
```
sprusr: Possibly some more comment to go with this (yeah sorry, I know I was the one who put this here!):

```js
// $FlowFixMe colonyJS types don't yet accept some methods as async
```

packages/colony-js-client/src/lib/EthersWrappedWallet/index.js
```diff
@@ -0,0 +1,145 @@
...
+import BigNumber from 'bn.js';
+import EthereumTx from 'ethereumjs-tx';
+import { utils } from 'ethers';
+import { isHex } from 'web3-utils';
+
+import type { ContractResponse } from '../../index';
+
+import type { TransactionOptions, TransactionRequest } from './types';
+
+type ContractResponseMeta = $PropertyType<ContractResponse<*>, 'meta'>;
```
sprusr: Can you put both of these in `types.js`?

packages/colony-js-client/src/getNetworkClient.js
ryanchristo: Hehe. Done.

packages/colony-js-client/src/lib/EthersWrappedWallet/index.js
ryanchristo: Done. 👍

chmanie:
Does this mean we can remove it from the dApp? Can we track that, maybe with an issue?

ryanchristo:
We would still need to export it from colonyJS to be used outside colonyJS. This was a quick fix to use purser with the new `getNetworkClient` method. @sprusr

JamesLefrere:
I can see why this has been done, and it's ok for now, but it would be best to align the interfaces and dependencies when we have the opportunity to do so.

JamesLefrere:
(and as Chris says, we should export that so we can simplify the dapp, too)

PR #389, opened 2019-04-29
DomainAuth sender

### LINKED ISSUE #386: Automatically determine method permission proofs
As of the horizontal modularity implementation in colonyNetwork, permissioned methods additionally accept `_permissionDomainId` and `_childSkillIndex` as arguments. These are used by the contract to determine whether the sender has permission in the relevant domain to perform the action they are attempting, without having to traverse the domain tree (expensive to do so).

So instead of the contract doing the tree traversal, colonyJS must do so instead! When trying to perform an action in a specific domain:

* Check whether the user has the permission in that domain `hasUserRole(walletAddress, domain, permission)`, if so then `_permissionDomainId = domain` and `_childSkillIndex = 0`
* If not then traverse up the tree with `getParentSkillId(childDomain, 0)`
* If the root domain is reached and permission is still not found, they don't have permission so throw

> ~TODO: is the above accurate?~ It is 😎

sprusr:
## Description

This PR adds the `DomainAuth` sender, which is used for methods which require a permission "proof" to be sent as arguments. This is the `_permissionDomainId` and `_childSkillIndex` seen in many colonyNetwork functions.

The `DomainAuth` sender works out what these values should be without any additional input from the user. Based on the `domainId` in which the user is trying to do something, and the `role` which they need in that domain, it will search the domain and all its parents until it finds a domain in which the user has permission - or if it fails to find it, the user does not have permission and the sender throws.

## TODO

- [x] Set the correct `role` for all senders in the `ColonyClient`
- [x] Refactor to allow for the case where there are multiple domains under a shared permission domain (`moveFundsBetweenPots`)

Closes #386

### DIFF
--- a/packages/colony-js-client/src/ColonyClient/index.js
+++ b/packages/colony-js-client/src/ColonyClient/index.js
@@ -1984,7 +1984,7 @@ export default class ColonyClient extends ContractClient {
         {
           address: async () => {
             const { address } = await this.getExtensionAddress.call({
-              contractName: 'OneTx',
+              contractName: 'OneTxPayment',
             });
             return address;
           },
--- a/packages/colony-js-client/src/ColonyClient/index.js
+++ b/packages/colony-js-client/src/ColonyClient/index.js
@@ -21,8 +21,7 @@ import AddPayment from './senders/AddPayment';
 import AddTask from './senders/AddTask';
 import MakePayment from './senders/MakePayment';
 import RemoveExtension from './senders/RemoveExtension';
-// eslint-disable-next-line no-unused-vars
-import DomainAuth from './senders/DomainAuth';
+import DomainAuth, { getDomainIdFromPot } from './senders/DomainAuth';
 import SetAdminRole from './senders/SetAdminRole';
 import SetFounderRole from './senders/SetFounderRole';
 import addRecoveryMethods from '../addRecoveryMethods';
@@ -257,8 +256,6 @@ export default class ColonyClient extends ContractClient {
   */
   addDomain: ColonyClient.Sender<
     {
-      permissionDomainId: number, // The ID of the domain in which the sender has permission.
-      childSkillIndex: number, // The index that the `domainId` is relative to the `permissionDomainId`.
       parentDomainId: number, // The ID of the parent domain.
     },
     {
@@ -328,8 +325,6 @@ export default class ColonyClient extends ContractClient {
   */
   addTask: ColonyClient.Sender<
     {
-      permissionDomainId: number, // The ID of the domain in which the sender has permission.
-      childSkillIndex: number, // The index that the `domainId` is relative to the `permissionDomainId`.
       specificationHash: IPFSHash, // The specification hash of the task (an IPFS hash).
       domainId: ?number, // The ID of the domain (default value of `1`).
       skillId: ?number, // The ID of the skill (default value of `null`).
@@ -540,8 +535,6 @@ export default class ColonyClient extends ContractClient {
   */
   finalizePayment: ColonyClient.Sender<
     {
-      permissionDomainId: number, // The ID of the domain in which the sender has permission.
-      childSkillIndex: number, // The index that the `domainId` is relative to the `permissionDomainId`.
       paymentId: number, // The ID of the payment.
     },
     {},
@@ -1065,10 +1058,6 @@ export default class ColonyClient extends ContractClient {
   */
   makePayment: ColonyClient.Sender<
     {
-      permissionDomainId: number, // The ID of the domain in which the sender has permission.
-      childSkillIndex: number, // The index that the `domainId` is relative to the `permissionDomainId`.
-      callerPermissionDomainId: number, // The ID of the domain in which the caller has permission.
-      callerChildSkillIndex: number, // The index that the `domainId` is relative to the `callerPermissionDomainId`.
       recipient: Address, // The address that will receive the payment.
       token: TokenAddress, // The address of the token contract (an empty address if Ether).
       amount: BigNumber, // The amount of tokens (or Ether) for the payment.
@@ -1130,9 +1119,6 @@ export default class ColonyClient extends ContractClient {
   */
   moveFundsBetweenPots: ColonyClient.Sender<
     {
-      permissionDomainId: number, // The ID of the domain that grants the address permission to call the method.
-      fromChildSkillIndex: number, // The index that the `domainId` is relative to the `permissionDomainId`.
-      toChildSkillIndex: number, // The index that the `domainId` is relative to the `permissionDomainId`.
       fromPot: number, // The ID of the pot from which funds will be moved.
       toPot: number, // The ID of the pot to which funds will be moved.
       amount: BigNumber, // The amount of funds that will be moved between pots.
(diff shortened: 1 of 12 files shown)


packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1989,6 +1980,26 @@ export default class ColonyClient extends ContractClient {
         domainId: DEFAULT_DOMAIN_ID,
         skillId: 0,
       },
+      permissions: [
+        {
+          address: async () => {
```
ryanchristo: Wow! Nice work here! Both the caller and the extension contract permissions with the optional `address` to override the default caller address. 👏

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1989,6 +1980,26 @@ export default class ColonyClient extends ContractClient {
...
+            const { address } = await this.getExtensionAddress.call({
+              contractName: 'OneTxPayment',
+            });
+            return address;
+          },
+          childSkillIndexNames: ['childSkillIndex'],
+          domainIds: ['domainId'],
+          permissionDomainIdName: 'permissionDomainId',
+          role: COLONY_ROLE_ADMINISTRATION, // TODO: also need to have COLONY_ROLE_FUNDING
```
ryanchristo: Did you mean to include this TODO in this PR?

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1989,6 +1980,26 @@ export default class ColonyClient extends ContractClient {
...
+        },
+        {
+          childSkillIndexNames: ['callerChildSkillIndex'],
+          domainIds: ['domainId'],
+          permissionDomainIdName: 'callerPermissionDomainId',
+          role: COLONY_ROLE_ADMINISTRATION, // TODO: also need to have COLONY_ROLE_FUNDING
```
ryanchristo: TODO here too... do?

packages/colony-js-client/src/ColonyClient/senders/DomainAuth.js
```diff
@@ -0,0 +1,270 @@
...
+  // name of the input value conatining the domainId, or function to get the domainId
+  domainIds:
+    | Array<string | ((inputValues: Object) => Promise<number>)>
+    | ((inputValues: Object) => Promise<number[]>),
+  // the name of the output value to set as the permisisonDomainId
+  permissionDomainIdName: string,
+  // the name of the output value to set as the childSkillIndex
+  childSkillIndexNames: string[],
+  // address for which the permission is required (default of current wallet address)
+  address?: string | (() => Promise<string>),
```
ryanchristo: 👏

packages/colony-js-client/src/ColonyClient/senders/DomainAuth.js
```diff
@@ -0,0 +1,270 @@
...
+import type { ColonyClient } from '../../index';
+
+type ColonyRole = $Keys<typeof COLONY_ROLES>;
+
+type PermissionType = {|
+  // the role required
+  role: ColonyRole,
+  // name of the input value conatining the domainId, or function to get the domainId
+  domainIds:
+    | Array<string | ((inputValues: Object) => Promise<number>)>
```
ryanchristo: 👏

packages/colony-js-client/src/ColonyClient/senders/DomainAuth.js
```diff
@@ -0,0 +1,270 @@
...
+      options,
+    );
+  }
+
+  /**
+   * For each required permission, get the permissionDomainId and
+   * childSkillIndex as proof of permission, under the specified input value
+   * keys. Return an object of input values.
+   */
+  async getPermissionProofs(inputValues: InputValues): Promise<Object> {
```
ryanchristo: 👏 👏 👏

packages/colony-js-contract-client/src/modules/paramTypes.js
```diff
@@ -57,7 +57,11 @@ const PARAM_TYPE_MAP: {
     },
     convertOutput(value) {
       if (!Array.isArray(value)) return [];
-      return value.map(element => (Number.isFinite(element) ? element : null));
+      return value.map(element => {
```
ryanchristo: Does this replace the need for `[bigNumber]`?

packages/colony-js-client/src/ColonyClient/index.js
sprusr: Yeah I did, since if you don't have that role as well as admin then you can't call this. But for the purposes of the dApp, I'm pretty sure if we set admin, we will also set funding (since using the old permissions extension).

This should be sorted out in the future, but didn't want to spend more time here while not strictly necessary.

packages/colony-js-client/src/ColonyClient/index.js
sprusr: Same here as above.

packages/colony-js-contract-client/src/modules/paramTypes.js
sprusr: It doesn't replace it, no. So my understanding of this type is it's for when we want the input/output to be an array of numbers (not BNs), but Ethers will give you back BNs regardless I think. So here we need to convert the output from being an array of BNs to an array of numbers.

The `[bigNumber]` type still gives you back an array of BNs.

sprusr:
As to whether `DomainAuth` should exist as another type of sender in `colony-js-contract-client`, no I don't think so. It's very specific to Colony contracts, and as such doesn't make much sense elsewhere.

And thanks for the spot on the changelog, will do 👍

packages/colony-js-client/src/ColonyClient/index.js
ryanchristo: Yup, probably not necessary for the dApp since it will be using OldRoles but would be nice to fix for developers using colonyJS in their own projects. If not in this PR, I think we should add support before a release.

packages/colony-js-contract-client/src/modules/paramTypes.js
ryanchristo: Did you run into this issue somewhere with a specific method? I trust your judgment but would like to better understand.

sprusr: Yep, I was getting this problem with all cases where the `[number]` type was used. For empty arrays it was fine, but if there were any values then all of them would be set to `null`, since each `element` was a BN, and so `Number.isFinite(element)` would return false.

Now, each `element` is first converted to a number (if it isn't already) before being checked.

ryanchristo:
Yea, that makes sense. Maybe not a sender class in colony-js-contract-client. I guess it is kinda like a helper rather than a sender because the sender methods use it but it's ok the way it is.

packages/colony-js-client/src/ColonyClient/index.js
sprusr: How do you feel about a beta release with this missing? Would like to get the dApp updated asap.

ryanchristo: Yea, a `pre` or `beta` release seems reasonable until this `TODO` is fixed and the docs are updated. Can you open an issue to track this task?

sprusr:
Tracking todos in #394

PR #392, opened 2019-05-01
Convert Ethers BNs to bn.js in paramTypes

### LINKED ISSUE #381: Convert Ethers flavoured BigNumbers to bn.js
Right now we're assuming that any `BigNumber`s Ethers returns are of the `bn.js` flavour. While they are based off this, they have a different set of methods available, which causes confusion - especially since our Flow types say everything is `bn.js` flavour!

A solution is to wrap all BN values we get back from Ethers with something like `new BigNumber(original.toString)`.

sprusr:
## Description

* Convert Ethers BNs to bn.js BNs in applicable paramType `convertOutput` functions

## TODO

- [ ] Should we proxy provider methods such as `getBalance` in the adapter so that we can convert BNs?

Closes #381

### DIFF
--- a/packages/colony-js-contract-client/src/__tests__/paramTypes.js
+++ b/packages/colony-js-contract-client/src/__tests__/paramTypes.js
@@ -61,7 +61,7 @@ describe('Parameter types', () => {
     expect(validateValueType(bn, 'bigNumber')).toBe(true);
 
     // Cleaning
-    expect(convertOutputValue(bn, 'bigNumber')).toBe(bn);
+    expect(convertOutputValue(bn, 'bigNumber')).toEqual(bn);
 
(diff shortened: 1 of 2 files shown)


chmanie:
Hmm. What should we do with this? @sprusr? Is this important?

sprusr:
Hello stranger! 👋

It’s not so important, just that Ethers has its own messed up version of `bn.js` so if you are referencing the docs from there it will likely not apply.

Might be redundant as of newer Ethers versions. But if not, at least types should be fixed to say messed up BN is returned.

chmanie:
Yeah I think this might be a bigger undertaking. Do you think we consistently return ethers BigNumbers otherwise (with this PR not merged)?

PR #399, opened 2019-05-07
Various fixes

sprusr:
## Description

* Return expected format object from GetExtensionAddress caller
* Custom estimation of DomainAuth sender
* Allow `0x0...` as an address

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Also, please note that colonyJS is currently using `ethers` version `3.0.27`.
 **Types**
 
 * Added `[number]` param type (`@colony/colony-js-contract-client`)
+* Renamed `tokenAddress` param type to `anyAddress` to better reflect its uses (`@colony/colony-js-contract-client`)
 
 **Maintenance**
 
--- a/packages/colony-js-client/scripts/docgen.js
+++ b/packages/colony-js-client/scripts/docgen.js
@@ -30,19 +30,19 @@ const CONTRACT_CLIENTS = [
 ];
 
 const TYPES = {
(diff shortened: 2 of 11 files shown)


sprusr:
Point for discussion: is allowing `0x0...` as an address a bad thing to do? I've done this since when you mint a token, it generates a `Transfer` event with the from address being `0x0...`. This was failing validation and breaking ERC 20 transfer event fetching if you had minted tokens.

ryanchristo:
If you want to allow `0x0`, it should be a separate validation, the same way we have type `Address` and type `TokenAddress` (which can be `0x0` for ethers). If it is not a separate validation, I believe it will break the role assignment and potentially other methods, which require checking for `0x0` as the set address for "no assigned role".

```
    makeExecuteTaskRoleChange('setTaskEvaluatorRole', async ({ taskId }) => {
      const { address } = await this.getTaskRole.call({
        taskId,
        role: TASK_ROLE_EVALUATOR,
      });
      if (isValidAddress(address))
        throw new Error('Unable to set task role; evaluator is already set');
      return null;
    });
```

sprusr:
Right, thanks for the guidance there @ryanchristo. I've used what was the `tokenAddress` type for those cases, but renamed to `anyAddress` to better reflect that it's used for other purposes now.

ryanchristo:
@sprusr Should we rename or remove `TokenAddress` as a type for doc generation as well?

ryanchristo:
```
type Address = string;
...
type TokenAddress = string;
```

```
Address: 'address',
...
TokenAddress: 'address',
```

PR #406, opened 2019-05-17
Support Goerli Infura provider in getNetworkClient

sprusr:
## Description

Does what it says! Since we're stuck on Ethers v3 for now, this borrows some `InfuraProvider` logic from v4 so that we can support connecting to the Goerli testnet.

### DIFF
--- a/packages/colony-js-client/src/getNetworkClient.js
+++ b/packages/colony-js-client/src/getNetworkClient.js
@@ -37,7 +37,7 @@ const getInfuraProvider = (network: string, infuraProjectId?: string) => {
 
   return new providers.JsonRpcProvider(
     `https://${host}/v3/${infuraProjectId || defaultInfuraProjectId}`,
-    network,
+    network === 'goerli' ? undefined : network,
   );
 };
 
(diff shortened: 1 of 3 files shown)


packages/colony-js-client/src/getNetworkClient.js
```diff
@@ -8,10 +8,47 @@ import { TrufflepigLoader } from '@colony/colony-js-contract-loader-http';
...
+    case 'kovan':
+      host = 'kovan.infura.io';
+      break;
+    default:
+      throw new Error(
+        `Could not get provider, unsupported network: ${network}`,
+      );
+  }
+
+  return new providers.JsonRpcProvider(
```
ryanchristo: This version of `JsonRpcProvider` does not accept `goerli` as a network. I think we can work around this by treating it like private network and doing something like:

```
  return new providers.JsonRpcProvider(
    `https://${host}/v3/${infuraProjectId || defaultInfuraProjectId}`,
    network === 'goerli' ? undefined : network,
  );
```

I was able to successfully test with the above solution.

PR #421, opened 2019-06-10
Upgrade for glider

sprusr:
## Description

This PR contains all the necessary changes to upgrade our support from `glider-rc.1` to `glider-rc.3`.

**New stuff** ✨

* Added the `ColonyRoleSet` event
* Added new ABIs to `NetworkLoader`, including old events
* Patch `lookupRegisteredENSDomain` for testnets (the contracts falsely return `.eth` domains when it should be `.test`) - see https://github.com/JoinColony/colonyNetwork/issues/645

**Changes** 🏗

* `ColonyRoleSet` is now set as what is emitted from role change methods - the old events remain in the client to allow for fetching old logs
* Docs types contract versions updated to `glider-rc.3`
* Better error for `DomainAuth` when user doesn't have permission
* Fix the `ContractClient` `getLogs` method to correctly format topic filters

**Deletions** ⚰️

* Removed Rinkeby deployments from the `NetworkLoader`, since we are no longer supporting these
* Removed unused networks from the `getNetworkClient` Infura providers

## TODO

- [x] Update `NetworkLoader`
- [x] Update docs types to `glider-rc.3` in `ColonyNetworkClient` (and other clients where necessary)
- [ ] ~Investigate handling of old logs when fetching new (proxy old events to new ones, something like that)~ Not urgent, maybe not needed
- [x] Fix log filters to work with Infura (is this still a problem?)


packages/colony-js-client/src/ColonyNetworkClient/callers/LookupRegisteredENSDomain.js
```diff
@@ -0,0 +1,28 @@
+/* @flow */
+
+import ContractClient from '@colony/colony-js-contract-client';
+
+export default class LookupRegisteredENSDomain extends ContractClient.Caller<
+  *,
```
thiagodelgado111: No defined types?

packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -138,26 +140,39 @@ export default class ContractClient {
...
+    });
 
     // Combine any existing topics with the extra ones
     if (!topics.length) {
       topics.push(extraTopics);
     } else if (Array.isArray(topics[0])) {
       topics[0].push(...extraTopics);
     } else {
-      topics[0] = [...topics[0], ...extraTopics];
+      topics[0] = [topics[0], ...extraTopics];
```
thiagodelgado111: 👍

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -89,6 +92,13 @@ type ColonyLabelRegistered = ContractClient.Event<{
 type ColonyRewardInverseSet = ContractClient.Event<{
   rewardInverse: BigNumber, // The reward inverse value that was set.
 }>;
+type ColonyRoleSet = ContractClient.Event<{
+  address: Address,
+  domainId: number,
+  role: ColonyRole,
+  setTo: boolean,
```
ryanchristo: Can we add comments for each parameter so that they have descriptions in the generated docs?

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -51,10 +51,12 @@ type TaskRole = $Keys<typeof TASK_ROLES>;
 type TaskStatus = $Keys<typeof TASK_STATUSES>;
 type AnyAddress = string;
 
+// @note Deprecated in glider-rc.3, replaced with ColonyRoleSet
```
ryanchristo: Nice to have this note while we explore our options for versioning 👍

packages/colony-js-client/src/TokenClient/index.js
```diff
@@ -79,7 +79,7 @@ export default class TokenClient extends ContractClient {
       contract: 'Token.sol',
       // eslint-disable-next-line max-len
       contractPath: 'https://github.com/JoinColony/colonyToken/blob/59cf56f18c54c0bc749ddaa8b6d77ebfd0d0aaf4/contracts',
-      version: 'glider-rc.1',
+      version: 'glider-rc.3',
```
ryanchristo: All colonyToken hashes need to be updated to `8e5139586018fbaa53e9cf971e040b3ae2a4bb92`.

packages/colony-js-client/src/ColonyClient/senders/DomainAuth.js
```diff
@@ -183,7 +183,10 @@ export default class DomainAuth<
             ),
           );
 
-          const highestDomain = await this.getHighestDomain(permissionDomains);
+          const highestDomain = await this.getHighestDomain(
+            permissionDomains,
+            role,
+          );
```
ryanchristo: 👏

packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -138,26 +140,39 @@ export default class ContractClient {
...
+   * events including that single topic, one of any of the array of topics, or
+   * any topic in that position respectively.
+   *
+   * - String: match only this topic in this position
+   * - Array: match any of these topics in this position
+   * - Null: match any topic in this position
+   *
+   * The returned logs will match these filters in each position. Trailing null
+   * values will require a topic in that position (e.g. [null, null] will only
+   * match logs with at least two topics).
```
ryanchristo: 👏

packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -253,18 +268,22 @@ export default class ContractClient {
...
+      Object.assign(this.events, {
+        [eventName]: event,
+      });
+
+      Object.assign(this.eventSignatures, {
+        [event.interface.topics[0]]: event,
+      });
+    } catch (error) {
+      console.info(error);
+    }
```
ryanchristo: Can you add a comment at the top of this try-catch block? This will allow for the initializing of a client even if the events do not match the events in the ABI, correct?

packages/colony-js-contract-loader-network/src/NetworkLoader.js
```diff
@@ -8,36 +8,33 @@ import ContractLoader, {
 import type { RequiredContractProps } from '@colony/colony-js-contract-loader';
 
 const NETWORKS = {
-  RINKEBY: 'rinkeby',
```
ryanchristo: 👏

packages/colony-js-client/src/ColonyClient/index.js
ryanchristo: Can you add a note to `ColonyRoleSet` that it was intoduced in `glider-rc.3`?

packages/colony-js-client/src/TokenClient/index.js
sprusr: Argh dammit yeah, great spot 💯

packages/colony-js-contract-client/src/classes/ContractClient.js
sprusr: Does this make sense to you? It's a complicated thing and I felt like I didn't do a great job explaining it

packages/colony-js-client/src/ColonyNetworkClient/callers/LookupRegisteredENSDomain.js
sprusr: Pretty sure the `*` should cause it to infer the types that are defined for it in `ColonyNetworkClient`. And if not internally, they are at least externally correct for people using colonyJS.

packages/colony-js-contract-client/src/classes/ContractClient.js
sprusr: Yep sorry totally should have a comment. That's right though, it's kinda hacky I feel like but yeah need some way to work when the old events aren't in the ABI (loading locally basically).

packages/colony-js-contract-client/src/classes/ContractClient.js
ryanchristo: It's a bit tricky to explain but it makes sense. I had to do some refreshing on logs and events but this seems ok to me.

PR #422, opened 2019-06-14
Update network loader to support mainnet

ryanchristo:
## Description

This pull request updates the network loader to support mainnet. This pull request also updates the `getNetworkClient` method to support a `mainnet` option.

**Changes** 🏗

* Added `mainnet` contract address to both versions of the `EtherRouter` contract.
* Updated the network loader to load contracts by version only. We no longer need to load contracts by network because `goerli` and `mainnet` are identical and `rinkeby` is no longer supported.
* Updated the network loader tests to include tests for both `goerli` and `mainnet`.


packages/colony-js-client/src/getNetworkClient.js
```diff
@@ -14,7 +14,7 @@ export const defaultInfuraProjectId = '7d0d81d0919f4f05b9ab6634be01ee73';
 const getInfuraProvider = (network: string, infuraProjectId?: string) => {
   let host;
   switch (network) {
-    case 'homestead':
+    case 'mainnet':
```
sprusr: Is `homestead` what others use? I think it's best to go with the crowd here. Maybe we should allow either, and also possibly default to this?

packages/colony-js-contract-loader-network/contracts/versioned/v1/EtherRouter.json
```diff
@@ -132,6 +132,9 @@
   "networks": {
     "goerli": {
       "address": "0x79073fc2117dD054FCEdaCad1E7018C9CbE3ec0B"
+    },
+    "mainnet": {
```
sprusr: Same with this, `homestead` or `mainnet` - I don't know which is technically correct, or even if you could say one is more correct than the other.

packages/colony-js-client/src/getNetworkClient.js
ryanchristo: Yea, probably safe to allow both. I was not entirely sure myself.

packages/colony-js-contract-loader-network/contracts/versioned/v1/EtherRouter.json
ryanchristo: I will add both just to be safe. 👍

rdig: I use `homestead` in `purser`: https://github.com/JoinColony/purser/blob/master/modules/node_modules/%40colony/purser-core/defaults.js#L126

But in the `purser` context it's a bit irrelevant as I only address networks by `id` and not by name

PR #428, opened 2019-07-02
Don't set nonce values for MetaMask (`EthersWrappedWallet`)

JamesLefrere:
## Description

This PR adjusts the `sendTransaction` method of `EthersWrappedWallet` such that it doesn't set a nonce value; MetaMask prefers to handle this.


ryanchristo:
@JamesLefrere Don't forget to update the changelog next time. 😉

PR #432, opened 2019-07-10
Add verbse option to ContractClient

### LINKED ISSUE #430: Better error reporting/info for events not found on ABI
## Description

Currently, when an event specified in the `ContractClient` instance is not found on the loaded ABI, an error is logged of the form: `No such event "x" on this contract`.

The info should instead inform something along the lines of "your ABI is different from what I expected, but I can continue anyway". Additionally, it would ideally not be logged when not in development build mode.

ryanchristo:
## Description

This pull request adds a `verbose` parameter to `ContractClient`. This provides the option to view additional logs such as `No such event "x" on this contract`.

**New stuff** ✨

* Add `verbose` option to `ContractClient`
* Add `verbose` option to `getColonyClient` method
* Add `verbose` option to `getNetworkClient` method

**Changes** 🏗

* Only log `No such event "x" on this contract` if `verbose` is `true`

Closes #430

### DIFF
--- a/packages/colony-js-client/src/ColonyClient/index.js
+++ b/packages/colony-js-client/src/ColonyClient/index.js
@@ -1918,12 +1918,13 @@ export default class ColonyClient extends ContractClient {
     query,
     tokenClient,
     tokenLockingClient,
+    verbose,
   }: {
     networkClient?: ColonyNetworkClient,
     tokenClient?: TokenClient,
     tokenLockingClient?: TokenLockingClient,
   } & ContractClientConstructorArgs) {
-    super({ adapter, query });
+    super({ adapter, query, verbose });
 
     if (!(networkClient instanceof ColonyNetworkClient))
       throw new Error(
@@ -1980,6 +1981,7 @@ export default class ColonyClient extends ContractClient {
     const tokenClient = new this.constructor.TokenClient({
       adapter: this.adapter,
       query: { contractAddress: tokenAddress },
+      verbose: this.verbose,
     });
     await tokenClient.init();
     return tokenClient;
--- a/packages/colony-js-contract-client/src/classes/ContractClient.js
+++ b/packages/colony-js-contract-client/src/classes/ContractClient.js
@@ -274,7 +274,7 @@ export default class ContractClient {
 
     // Allow initialising of clients where some events may be missing in the
     // ABI, due to changing of events on the contract and then log the error
-    // if the client is initialized in verbose mode.
+    // as a warning if the client is initialized in verbose mode.
     try {
       const event = new ContractEvent({
         eventName,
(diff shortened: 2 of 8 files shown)


packages/colony-js-contract-client/src/classes/ContractClient.js
```diff
@@ -285,7 +290,9 @@ export default class ContractClient {
         [event.interface.topics[0]]: event,
       });
     } catch (error) {
-      console.info(error);
+      if (this.verbose) {
+        console.info(error);
```
sprusr: Part of this PR should also be to change the message that is logged to be more descriptive (see issue).

sprusr: I think it should probably mention that it's not found in the loaded ABI or something. Didn't mean for those words exactly! :)

ryanchristo: Haha. I realize that. What about listing the colony contract version before the warning messages by logging the version in the init if verbose is true?

```
The colony client you are initializing is using Colony.sol version 2
WARNING: No such event "ColonyAdministrationRoleSet" on this contract
WARNING: No such event "ColonyArchitectureRoleSet" on this contract
WARNING: No such event "ColonyFundingRoleSet" on this contract
WARNING: No such event "ColonyRootRoleSet" on this contract
```

ryanchristo: Actually, not sure if it's possible to list the version before these warnings.

ryanchristo:
```
WARNING: Some events and methods might not be available to the client based on the ABI loaded for this colony contract version.
WARNING: No such event "ColonyAdministrationRoleSet" on this contract
WARNING: No such event "ColonyArchitectureRoleSet" on this contract
WARNING: No such event "ColonyFundingRoleSet" on this contract
WARNING: No such event "ColonyRootRoleSet" on this contract
```

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1918,12 +1918,20 @@ export default class ColonyClient extends ContractClient {
...
     networkClient?: ColonyNetworkClient,
     tokenClient?: TokenClient,
     tokenLockingClient?: TokenLockingClient,
   } & ContractClientConstructorArgs) {
-    super({ adapter, query });
+    super({ adapter, query, verbose });
+
+    if (verbose) {
+      console.warn(
+        'WARNING: Some events and methods might not be available based on ' +
```
sprusr: Do you think this is useful to unconditionally log when we are in verbose mode? An improvement might be only when we load from a local source. But I think better would be just provide this context with the error itself.

ryanchristo: Not necessarily. I was just trying to provide more context. I'm good with removing this and just using your recommended warning message: `No such event "x" in loaded ABI`.

PR #436, opened 2019-07-23
Add custom Infura provider

sprusr:
## Description

This provider retries queries when a bad (`0x`) response is received. We should test this and ideally reproduce the error and see that this sorts it before merging.

**Changes** 🏗

* Use `InfuraProvider` in place of `getInfuraProvider`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## NEXT
+
+**Bug fixes**
+
+* Add custom Infura provider which retries queries when bad responses are recieved, which happens occasionally (`@colony/colony-js-client`)
+
 ## v1.13.3
 
 **Bug fixes**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Bug fixes**
 
-* Add custom Infura provider which retries queries when bad responses are recieved, which happens occasionally (`@colony/colony-js-client`)
+* Add custom Infura provider which retries queries when bad responses are received, which happens occasionally (`@colony/colony-js-client`)
 
 ## v1.13.3
 
--- a/packages/colony-js-client/src/getNetworkClient.js
+++ b/packages/colony-js-client/src/getNetworkClient.js
@@ -9,68 +9,7 @@ import NetworkLoader from '@colony/colony-js-contract-loader-network';
 import { TrufflepigLoader } from '@colony/colony-js-contract-loader-http';
 import ColonyNetworkClient from './ColonyNetworkClient/index';
 import EthersWrappedWallet from './lib/EthersWrappedWallet/index';
-
-export const defaultInfuraProjectId = '7d0d81d0919f4f05b9ab6634be01ee73';
-
-// Custom provider for use with Infura which retries query requests if the
-// result is unexpected data.
-// Related issue: https://github.com/INFURA/infura/issues/157
-class InfuraProvider extends providers.JsonRpcProvider {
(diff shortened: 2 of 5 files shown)


sprusr:
Okay so the thing is, a response of `0x` can also be valid. For example if you attempt to call a function that doesn't exist on a contract, I'm pretty sure that returns `0x`.

When I encountered the bug that we're trying to fix, I retried the request straight after and it worked, so I'm thinking once should be enough. But we really need to reproduce this to be sure.

ryanchristo:
> When I encountered the bug that we're trying to fix, I retried the request straight after and it worked, so I'm thinking once should be enough.

The issue stated one or two times... I'm worried about "should be enough" but if you feel confident that one is enough that one is enough.

> I'm pretty sure that returns 0x

We should probably confirm this. It would be nice to have a specific error (...and a retry).

packages/colony-js-client/src/getNetworkClient.js
```diff
@@ -12,34 +12,65 @@ import EthersWrappedWallet from './lib/EthersWrappedWallet/index';
...
+      result === '0x'
+    ) {
+      if (this._verbose) {
+        console.info(`Bad Infura response for method "${method}"; retrying.`, {
+          method,
+          params,
+          result,
+        });
+      }
+      return super.perform(method, params);
```
JamesLefrere: I wonder if it might be prudent to add a delay and/or retry mechanism here, if we're waiting for a change.

ryanchristo:
I was able to reproduce locally but unfortunately it did not resolve the issue.

Screen Shot 2019-07-24 at 7 29 58 PM

ryanchristo:
Hmmm.... Another test to add. This time it worked successfully two times in a row.

Screen Shot 2019-07-24 at 7 44 36 PM

ryanchristo:
Success but threw the error... 🤔

Screen Shot 2019-07-24 at 8 53 25 PM

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## NEXT
+
+**Bug fixes**
+
+* Add custom Infura provider which retries queries when bad responses are recieved, which happens occasionally (`@colony/colony-js-client`)
```
JamesLefrere: typo: received

packages/colony-js-client/src/lib/InfuraProvider/index.js
```diff
@@ -0,0 +1,94 @@
...
+    ) {
+      // Set the number of retries. The time between retries will increase with
+      // a default exponential factor of 2.
+      const retries = 7;
+
+      // Retry if the result is '0x'. '0x' is an acceptable response if the
+      // method does not exist on the contract, therefore, the promise must
+      // eventually resolve with `0x` after the given number of retries.
+      return promiseRetry({ retries }, (retry, number) => {
+        if (this._verbose) {
```
sprusr: Won't this just happen for every request, even the first try?

ryanchristo: If the result from the first try does not meet the conditions that this is within (ie, if the result is not `0x`), it will not happen on the first try and it will return the result.

sprusr: Oh yeah damn, add another to the heat excuse counter

ryanchristo: Haha. No worries! We did this late last night so good that you double checked. 👍

PR #439, opened 2019-07-29
Use pending block tag to tell bad Infura response

### LINKED ISSUE #437: Improve experimental InfuraProvider
## Description

Currently we are retrying a call if the result is `0x` as a workaround to an issue where Infura will sometimes return `0x` as a result when it should return a valid result. The retry solution solved the problem but it also introduced a minor issue where `0x` is the expected result and now we have to wait for the retries to finish before receiving the expected result.

We think the unexpected `0x` response is happening when a more complex contract (ie, a colony contract) is created and additional calls to interact with that contract are called immediately after and attempting to communicate with nodes that have not been updated with the latest block.

## Suggestions

If the unexpected `0x` result is the result of calls to nodes that have not been updated, we should check the latest block number before attempting retries in the custom `InfuraProvider` that was introduced in #436.

Checking the latest block number might shed light on the Infura issue and it might also provide a solution where we would only attempt retries if we were not using the latest block number.

sprusr:
## Description

We've worked out that where a particular `eth_call` request gets `0x` back when it shouldn't, an equivalent call with the block tag of `pending` (default is `latest`) will give us the correct result.

Now we are using this to determine if the initial `0x` response is invalid, and then waiting for the call with `latest` block tag to succeed. This is to make sure that the transactions we're using for the state are confirmed.

**Changes** 🏗

* `InfuraProvider` uses `eth_call` with block tag of `pending` to determine if a `0x` response was invalid, before retrying. This prevents retries in situations where they are not needed.

## TODO

- [ ] Some more testing before approving please!

Resolves #437

### DIFF
--- a/packages/colony-js-client/src/lib/InfuraProvider/index.js
+++ b/packages/colony-js-client/src/lib/InfuraProvider/index.js
@@ -3,8 +3,6 @@
 import { providers } from 'ethers';
 import promiseRetry from 'promise-retry';
 
-import type { Params } from './types';
-
 export const defaultInfuraProjectId = '7d0d81d0919f4f05b9ab6634be01ee73';
 
 // Custom provider for use with Infura which retries query requests if the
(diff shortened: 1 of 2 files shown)


packages/colony-js-client/src/lib/InfuraProvider/index.js
```diff
@@ -46,49 +44,55 @@ export default class InfuraProvider extends providers.JsonRpcProvider {
...
-    if (
-      (method === 'call' ||
-        method === 'estimateGas' ||
-        method.startsWith('get')) &&
-      result === '0x'
-    ) {
-      // Set the number of retries. The time between retries will increase with
-      // a default exponential factor of 2.
-      const retries = 7;
+  async perform(method: string, params: *) {
```
ryanchristo: Why remove `Params` type? Should we update the `Params` type or provide two distinct types in order to account for the `pending` call rather than simply removing it?

packages/colony-js-client/src/lib/InfuraProvider/index.js
```diff
@@ -46,49 +44,55 @@ export default class InfuraProvider extends providers.JsonRpcProvider {
...
+    if (method === 'call') {
```
ryanchristo: Why remove `estimateGas` and methods starting with `get`?

We are currently calling `estimateGas` on every call in colonyJS when using the EthersWrappedWallet. I have a feeling the "always failing transaction" error might be related to this issue on occasion.

packages/colony-js-client/src/lib/InfuraProvider/index.js
sprusr: Ah oops, yeah I'll add it back. Basically there's a whole ton of stuff that can possibly be in there, but I suppose we're only concerned with what we'll use.

packages/colony-js-client/src/lib/InfuraProvider/index.js
sprusr: Hm yeah should probably account for various other methods too, you're right. I removed because we only seem to experience this with `eth_call`, but if you think there's another then that's definitely reason enough to add back.

Size of this is going to grow somewhat though 😬

packages/colony-js-client/src/lib/InfuraProvider/index.js
ryanchristo: > I suppose we're only concerned with what we'll use.

👍

packages/colony-js-client/src/lib/InfuraProvider/index.js
ryanchristo: Yea.... might be good to add a comment as well explaining that we are not sure the extent of the issue but we definitely need this for `call` and maybe `estimateGas`.

Given the number of "always failing transaction" errors I've received and sometimes not a clear reason why, I think `estimateGas` should be in there at the very least.

PR #440, opened 2019-08-02
Add `eth_call` support and revert reason strings

### LINKED ISSUE #429: Get revert reason strings for failed transactions/estimations
## Description

When estimating gas for or sending a transaction, if that transaction fails I would like to have the revert reason string returned.

When connected to Ganache testnets, these errors are returned via RPC as the response to the estimate or send call. However, this is not the case for geth or parity nodes. In these cases, we can get the revert error (with reason string) by calling (as constant; `call`) the contract functions.

A PR which solves this issue should:
- [x] On estimate or send error, call the same function
- [x] Parse the returned error to extract the reason string
- [x] Return the reason string as a property of the thrown error

It may be possible to do this without first upgrading to Ethers 4.0, but upgrading will certainly make this easier (calling non-constant functions).

JamesLefrere:
## Description

This PR adds support for `eth_call` via `ContractMethodSender.call()`, so that transactions can be simulated. Additionally, revert reason strings are now parsed when calling `call`, `estimate` and `send` for transactions.

**New stuff** ✨

* Add support for `eth_call` for Senders in order to simulate sending the transaction and receive a result or parsed revert error
* Add `reason` property to transaction receipts
* Add `Adapter.callTransaction` method

**Changes** 🏗

* Rename `Contract.callTransaction` to `Contract.sendTransaction` for clarification

Resolves #429

### DIFF
--- a/packages/colony-js-adapter-ethers/src/EthersAdapter.js
+++ b/packages/colony-js-adapter-ethers/src/EthersAdapter.js
@@ -9,6 +9,7 @@ import type {
   IWallet,
   Signature,
   Transaction,
+  TransactionRequest,
 } from '@colony/colony-js-adapter';
 import type { Query } from '@colony/colony-js-contract-loader';
 
@@ -18,6 +19,9 @@ import type { ConstructorArgs } from './flowtypes';
 import EthersContract from './EthersContract';
 
 import { DEFAULT_TRANSACTION_WAIT_TIMEOUT } from './defaults';
+
+// See https://solidity.readthedocs.io/en/v0.4.24/control-structures.html#error-handling-assert-require-revert-and-exceptions
+const SOLIDITY_ERROR = '0x08c379a0';
 
 export default class EthersAdapter implements IAdapter {
   loader: ContractLoader;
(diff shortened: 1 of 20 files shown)


packages/colony-js-adapter-ethers/src/EthersAdapter.js
```diff
@@ -32,6 +32,27 @@ export default class EthersAdapter implements IAdapter {
     this.wallet = wallet;
   }
 
+  async callTransaction(transaction: $Shape<Transaction>): Promise<string> {
```
sprusr: This type is a bit vague. I think only these should be in it (and each optional):
```
from
to
gasLimit
gasPrice
value
data
```

packages/colony-js-contract-client/src/classes/ContractMethodSender.js
```diff
@@ -73,6 +99,16 @@ export default class ContractMethodSender<
     );
     const eventData = this.client.getReceiptEventData(receipt);
 
+    // If the transaction failed, call it directly and add the revert
```
sprusr: Wow! Fantastic!

packages/colony-js-adapter-ethers/src/EthersAdapter.js
JamesLefrere: Good call, I'll make that change :+1:

packages/colony-js-contract-client/src/classes/ContractMethodSender.js
```diff
@@ -73,6 +99,24 @@ export default class ContractMethodSender<
...
+    // reason to the receipt.
+    if (receipt && receipt.status === 0) {
```
JamesLefrere: PS missed off a check for `receipt` here, in now.

packages/colony-js-contract-client/src/classes/ContractMethodSender.js
```diff
@@ -73,6 +99,24 @@ export default class ContractMethodSender<
...
+      try {
+        const { from, to, data, gasPrice, gasLimit, value } = transaction;
+        await this.client.callTransaction({
+          data,
+          from,
+          gasLimit,
+          gasPrice,
+          to: to || '',
```
JamesLefrere: `to` can be `null`

PR #443, opened 2019-10-10
Add explanation transaction to unlock token

area:
Based on [this post](https://build.colony.io/t/colonyclient-issue-while-connecting/278/11) in the forums, it looks like dealing with the locked token wasn't part of the tutorial?


chmanie:
This looks like it can be safely merged!

PR #447, opened 2019-11-20
Burgundy Glider Update

area:
Before we can release a new version of the contracts, ColonyJS needs to know how to talk to them! That's what this PR is... broadly speaking, this comes down to letting it interact correctly with expenditures. The only wrinkle here is that on the contracts, we have two functions that are named `transferExpenditure`. I have added a custom sender that establishes which to used based on whether the address is the owner of the expenditure or not, which assumes that if you are the owner of the expenditure, you wish to use the corresponding function, and will only use the Arbitration permission if you are not the owner.

`transferExpenditureViaOwnership` and `transferExpenditureViaArbitration` are also available for the case that someone implementing a process via ColonyJS wishes to explicitly control which mechanism is used.

This is pretty far outside my wheel house, so fully expecting there to be a lot of improvements. I've also not touched the tests at all...


packages/colony-js-client/src/ColonyClient/senders/AddExpenditure.js
```diff
@@ -0,0 +1,20 @@
+/* @flow */
+/* eslint-disable import/no-cycle */
+
+import DomainAuth from './DomainAuth';
+
+// XXX This is a good use-case for some kind of async validation step,
+// but since the underlying method functionality is due to change very soon,
+// we're opting to not make big changes to the Sender behaviour, and simply
+// extend the `send` method to perform this async validation.
```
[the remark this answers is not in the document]
area: Heh, sadly this is copy-and-pasted from an existing file, so I can't claim credit for it ;-)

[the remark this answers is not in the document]
area: Heh, sadly this is copy-and-pasted from an existing file, so I can't claim credit for it ;-)

chmanie:
I rebased this. @area is this still up-to-date? Do you need to add anything?

area:
Yup, comparing to `develop` in colonyNetwork, I believe this is still up to date.

packages/colony-js-client/src/ColonyClient/index.js
```diff
@@ -1887,6 +2179,57 @@ export default class ColonyClient extends ContractClient {
...
+    ColonyClient,
+    {
+      contract: 'ColonyExpenditure.sol',
+      interface: 'IColony.sol',
+      version: 'burgundy-glider',
+    },
+  >;
+
+  /*
+  Transfer Expenditure as via the arbitration permission
```
kronosapiens: This comment seems incorrect?

packages/colony-js-client/src/ColonyNetworkClient/index.js
```diff
@@ -17,6 +17,7 @@ import CreateToken from './senders/CreateToken';
 import addRecoveryMethods from '../addRecoveryMethods';
 
 type Address = string;
+type AnyAddress = string;
```
kronosapiens: What is this type, above and beyond just a regular `Address`?

chmanie: I feel like they used to be different once

packages/colony-js-client/src/ColonyClient/index.js
chmanie: I'll make a PR with some fixes and will include this

PR #450, opened 2020-01-21
Replace ARCHITECTURE_SUBDOMAIN with ARCHITECTURE

kronosapiens:
Replace every reference to `ARCHITECTURE_SUBDOMAIN` with `ARCHITECTURE`, update the docs.


packages/colony-js-client/src/ColonyClient/senders/DomainAuth.js
```diff
@@ -329,7 +329,7 @@ export default class DomainAuth<
     });
 
     // Architecture can only architect child domains
-    if (hasDomainPermission && role !== COLONY_ROLE_ARCHITECTURE_SUBDOMAIN) {
+    if (hasDomainPermission && role !== COLONY_ROLE_ARCHITECTURE) {
```
area: Is this right? Unlike `ARCHITECTURE_SUBDOMAIN`, there are actions in domain `x` I can take if I have been assigned `ARCHITECTURE` in domain `x` (creating a domain).

kronosapiens: Yeah I had noticed that. If on the network side we instead enforce the restriction through a modifier (rather than a role), we may have to add a list of the restricted function here as well.

kronosapiens: Ok, reading through this function more and there's a good bit of weirdness here. Let me see if I can sort through it properly.

kronosapiens: Ok, so I think this is fine for the following: currently, we have only two layers of domains, meaning that architecture permission is currently inert in child domains. This is because they cannot change roles in the domain of the permission, and further they cannot create new domains because we don't support a third level of domains. So for now it's ok that we treat the architecture role like this. Once we allow multiple levels of domain, this logic will need to be re-worked.

kronosapiens:
Ha well we should never be committing whitespace anyway, so this is penance for some past omission.

PR #458, opened 2020-02-18
Update handling of colony creation & upgrading

kronosapiens:
- [x] Introduce a `MAX_VERSION` constant indicating the highest Colony version supported.
- [x] Create a custom sender for the `upgrade` function, which checks against `MAX_VERSION`.
- [x] Use the overloaded `createColony` with no-op defaults.
- [x] Handle the case of a user initiating a `ColonyClient` for an unsupported colony.

### DIFF
--- a/packages/colony-js-client/src/ColonyNetworkClient/index.js
+++ b/packages/colony-js-client/src/ColonyNetworkClient/index.js
@@ -14,6 +14,7 @@ import TokenLockingClient from '../TokenLockingClient/index';
 
 import LookupRegisteredENSDomain from './callers/LookupRegisteredENSDomain';
 import CreateToken from './senders/CreateToken';
+import CreateColony from './senders/CreateColony';
 import addRecoveryMethods from '../addRecoveryMethods';
 
 type Address = string;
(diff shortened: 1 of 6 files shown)


packages/colony-js-client/src/ColonyNetworkClient/index.js
```diff
@@ -1103,6 +1109,25 @@ export default class ColonyNetworkClient extends ContractClient {
...
+      functionName: 'createColony',
+      input: [
+        ['tokenAddress', 'address'],
+        ['version', 'number'],
+        ['colonyName', 'string'],
+        ['orbitdb', 'string'],
+        ['useExtensionManager', 'boolean'],
+      ],
+      defaultValues: {
+        version: 0,
```
chmanie: Should this be `MAX_VERSION` maybe?

packages/colony-js-client/src/constants.js
```diff
@@ -1,5 +1,7 @@
 /* @flow */
 
+export const MAX_VERSION = 3; // Auburn glider
```
chmanie: I think this should be `4`, shouldn't it? Given we have merged this: https://github.com/JoinColony/colonyJS/pull/447

packages/colony-js-client/src/ColonyNetworkClient/index.js
```diff
@@ -194,17 +195,22 @@ export default class ColonyNetworkClient extends ContractClient {
...
       ColonyAdded: ColonyAdded,
       RecoveryRoleSet: RecoveryRoleSet,
+      ColonyLabelRegistered: ColonyLabelRegistered,
     },
     ColonyNetworkClient,
     {
       contract: 'ColonyNetwork.sol',
       interface: 'IColonyNetwork.sol',
-      version: 'glider',
+      version: 'burgandy-glider',
```
chmanie: I guess this is a typo

packages/colony-js-client/src/ColonyNetworkClient/index.js
kronosapiens: Mmm... yes you're right. I was thinking `0` since the network will interpret that as the "current version", but I suppose that we don't want ColonyJS to create a colony at "current version" if it's greater than "max version". So yes, I'll update.

packages/colony-js-client/src/constants.js
kronosapiens: Yup! Will update 🙇

kronosapiens:
Ok nearly there but something weird is happening:

https://joincolony.slack.com/archives/C02R9SP3J/p1583264120059500

PR #461, opened 2020-02-25
Latest version is 3 until the app is updated

area:
There was a [change to an event](https://github.com/JoinColony/colonyNetwork/pull/693#discussion_r341070143) I had forgotten about, so version 3 colonies are not transparently supported by the v4 interfaces. As a result, for now, let's make colonyJS forget v4 contracts exist.


area:
Chris [beat me to it](https://github.com/JoinColony/colonyJS/commit/34feb8d692f503f396fcd611139ecffbb4ac92cf)

PR #464, opened 2020-03-31
Update colony-js-contract-loader-network to v4

kronosapiens:
## Description

Update the contract ABIs for `burgundy-glider` and bump the `CURRENT_VERSION` to `4`.


kronosapiens:
@chmanie Out of curiosity, what attributes of the ABI are needed for ColonyJS? Specifically I'm wondering if the `networks` and `bytecode` attributes are necessary, since some ABIs here have them and others do not.

PR #477, opened 2020-11-10
Fetch all Colony Members Reputation

rdig:
## Description

This PR adds in the `getMembersReputation` common extension, that fetches the list of all members who have reputation in the colony, sorted by said reputation.

While I was at it, I've also removed the strict network check for `getReputation`.

**Changes**

- [x] Add `getMembersReputation` method and types to common extensions
- [x] Remove `getReputation` strict network check


area (approved):
I have no more 'good ideas'!

PR #478, opened 2020-11-23
Add RecoveryRoleSet event interface to v3

rdig:
This PR manually adds in the `RecoveryRoleSet` event interface to the `v3` contracts _(as opposed to generating them automatically)_

This was needed, since the `awkwardRecoveryRoleEventClient` did not provide sufficient safeguard for `v3` colonies and so, fetching the `RecoveryRoleSet` events in the [helpers](https://github.com/JoinColony/colonyJS/blob/815b66fef7f484aa792ffa3557358ef657eae274/src/helpers.ts#L255) returns an array containing a `null` value. Eg:

```js
[null]
```

The helpers then [try to deconstruct](https://github.com/JoinColony/colonyJS/blob/815b66fef7f484aa792ffa3557358ef657eae274/src/helpers.ts#L282) a `values` object from this null value, which, in turn, turns into an error that crashes both colonyJS and the app.

See the error in the app:
Screenshot from 2020-11-23 17-02-45

_(The above is actually coming from `apollo` since when the `helpers` throw the respective error, we're actually inside the `ColonyFromName` query)_

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js",
-  "version": "3.0.0-alpha.5",
+  "version": "3.0.0-alpha.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js",
-  "version": "3.0.0-alpha.7",
+  "version": "3.0.0-alpha.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js",
-  "version": "3.0.0-alpha.6",
+  "version": "3.0.0-alpha.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js",
-  "version": "3.0.0-alpha.7",
+  "version": "3.0.0-alpha.8",
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
   "license": "GPL-3.0-only",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js",
-  "version": "3.0.0-alpha.6",
+  "version": "3.0.0-alpha.7",
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
   "license": "GPL-3.0-only",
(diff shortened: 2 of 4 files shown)


src/contracts/3/IColony.d.ts
```diff
@@ -503,6 +503,10 @@ interface IColonyInterface extends Interface {
   };
 
   events: {
+    RecoveryRoleSet: TypedEventDescription<{
```
chmanie, line 506: Was this in the contract all along? Why wasn't it automatically generated in the first place?

rdig: That's the thing. This is what broke the colonies on `goerli` with the new version of `colonyJS` deployed.

@area tells me that `RecoveryRoleSet` was emitted all the way back from v1.

That's why I feel uneasy about having to add these values manually, rather than have them generated.

chmanie: You should _never_ touch anything in the `contracts` directory by yourself, these are all autogenerated and will be overwritten when an update is necessary (for example of TypeChain). So It seems we need to find a different solution here

chmanie: I don't really understand though. The events were emitted but not defined in the ABI. Is that even possible @area? If so, can we add them to the actual contract and make a new tag (and release in the network repo) that we can use to autogenerate the ABI and types?

area: > The events were emitted but not defined in the ABI.
Yes
> Is that even possible @area?
Yes

This was what was meant to be dealt with by [this](https://github.com/JoinColony/colonyJS/blob/af6853c30c108f2ab21c68cd416525791e262e2b/src/clients/Colony/extensions/commonExtensions.ts#L953) is my understanding, but it doesn't work, which lead Raul down this road.

chmanie: OK, well, then I guess the way to do it is to repair the actual solution. Maybe we can find the reason why the `awkwardRecoveryRoleEventClient` does not have these events defined? They're definitely in the ABI.

chmanie: Ah. Got it.
This [line](https://github.com/JoinColony/colonyJS/blob/815b66fef7f484aa792ffa3557358ef657eae274/src/helpers.ts#L255)

should be
```
const recoveryRoleEvents = await getEvents(client.awkwardRecoveryRoleEventClient, recoveryRoleSetFilter);
```

That's the only thing that needs changing IMHO.

rdig:
@chmanie @area 

Chris was right, that change did it. It's live currently on QA, so you can check it out.

I had to change the types around about to keep TS happy. Please have a look and let me know if you have any changes regarding those.

After that, I'll merge it.

chmanie:
Yeah the types are not really correct like this. I'll take a look

rdig:
Yeah, that's what I figured. Thanks Chris! :+1:

PR #482, opened 2021-03-11
Introduce Extension Versioning

rdig:
This PR introduces versioning logic for extensions so that they can be installed at a certain version and upgraded to the newest one.

While at it, this PR also introduces a new extension: Voting Reputation

Resolves DEV-257


rdig:
Note: this branch / feature is to be considered "done"

It will be kept open and un-merged until the network side does a release for `DandelionLightweightSpaceship`. Until then it will be kept up-to-date _(for the most part)_ with colonyNetwork's `develop` from which various release candidates will be published.

PR #483, opened 2021-03-24
Fix setUserRolesWithProofs method

alicjakujawa:
This PR fixes setUserRolesWithProofs method. 

Initial implementation was passing Architecture Role to getPermissionProofs method for all domains. To change any permission in root domain we need Root permission, not Architecture permission. 

Adding here check if we're in root domain and passing correct role depend on this


rdig (approved):
This looks good to me.

After you merge this, do you want me to do a minor release so that you can use in the dapp? Or do you want to wait until version 4 comes out ?

area:
Tacit post-hoc approval from me, too :sweat_smile:

PR #485, opened 2021-04-05
Add customRootHash param to getReputation method

ArmandoGraterol:
In cases where you need reputation proof to do things on a motion(like staking), you need to use the root hash provided in the `motion` data, which may not be the current root hash. Otherwise, the values provided will be wrong.

Contributes to DEV-256

### DIFF
--- a/src/clients/Colony/extensions/commonExtensions.ts
+++ b/src/clients/Colony/extensions/commonExtensions.ts
@@ -198,6 +198,7 @@ export type ExtendedIColony<T extends AnyIColony = AnyIColony> = T & {
   getReputation(
     skillId: BigNumberish,
     address: string,
+    customRootHash?: string,
   ): Promise<ReputationOracleResponse>;
 
   getMembersReputation(
@@ -865,6 +866,7 @@ async function getReputation(
   this: ExtendedIColony,
   skillId: BigNumberish,
   address: string,
+  customRootHash?: string,
 ): Promise<ReputationOracleResponse> {
   if (!isAddress(address)) {
     throw new Error('Please provide a valid address');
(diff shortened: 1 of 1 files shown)


src/clients/Colony/extensions/commonExtensions.ts
```diff
@@ -874,7 +876,7 @@ async function getReputation(
 
   const skillIdString = bigNumberify(skillId).toString();
 
-  const rootHash = await this.networkClient.getReputationRootHash();
+  const rootHash = customRootHash ? customRootHash : await this.networkClient.getReputationRootHash();
```
rdig, line 879: Why is this change needed? Do you need to pass the root hash for a different domain? 

Can you provide an usage example?

ArmandoGraterol: While I was working on the `stakeMotion` saga, I encountered an error: `voting-rep-invalid-root-hash`. Consulted with Alex for more info, he told that's what happens when the `key/value/branchmask/siblings` I provided are wrong, and I should check that I'm providing proof from the reputation root hash in the `motion.rootHash`, and not necessarily the current root hash. He commented that `getReputation` always uses the current reputation root hash, which is not going to work.

If this change is introduced, I would use it like this in the `stakeMotion` saga, to get the correct values:

```
    const { skillId } = yield call(
      [colonyClient, colonyClient.getDomain],
      motionDomainId,
    );

    const { key, value, branchMask, siblings } = yield call(
      colonyClient.getReputation,
      skillId,
      userAddress,
      motionRootHash
    );

```

rdig: Where's the `motionRootHash` value coming from ?

ArmandoGraterol: This is also something that will most likely be used in the `objection` saga if I had to guess

ArmandoGraterol: > Where's the `motionRootHash` value coming from ?

It's coming from the motion data we receive, like this `motion.rootHash` (right now I'm pulling it from the events, just like the motion's id, and motion's domain id)

rdig (approved):
Good to go.

Let me know after you merge it so that I can do a release for you.

PR #507, opened 2022-02-17
Add compiled contracts for all network versions

chmanie:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

(Resolves | Contributes to) #31415


src/abis/dlwss/Version3.json
```diff
@@ -0,0 +1 @@
+{"contractName":"Version3","abi":[{"inputs":[],"name":"version","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"pure","type":"function"}],"bytecode":"0x6080604052348015600f57600080fd5b5060808061001e6000396000f3fe6080604052348015600f57600080fd5b506004361060285760003560e01c806354fd4d5014602d575b600080fd5b60336045565b60408051918252519081900360200190f35b60039056fea2646970667358221220113903f209a2543ad0df26aeaf9669f1fa7b3b23727c45de1f8fc5467f3729d064736f6c63430007030033"}
```
rdig: This _(as well as version `4`)_ is very confusing for a contract name. I also noticed it repeats over the various releases.

I don't think we can actually do something about it, but might as well be aware that there's potential for confusion here.

rdig (approved):
I'm good with this as long as we don't directly export the abi's from the library so as not to make the package footprint bigger when building for production.

src/abis/dlwss/Version3.json
chmanie: Yes, that's true. The ABIs are more thought of as a reference, so to be able to quickly re-build old contract clients on a new Typechain/ethers version. They also won't be checked in the npm package.

PR #508, opened 2022-02-23
Maint/fix added abis

chmanie:
## Description

The old PR (#507) missed some ABIs (and added some unnecessary ones) that were supposed to be for the `colonyToken`. This PR fixes that.

**New stuff** ✨

* Add some token ABIs that weren't there before


rdig (approved):
Good to go

PR #509, opened 2022-02-23
Rework build workflow

chmanie:
## Description

This PR changes how the typechain types are generated from the contract ABIs. They are now better organized by contract names:

```
CoinMachine/
IColony/
OneTxPayment/
VotingReputation/
Whitelist/
factories/
IColonyNetwork.d.ts
Token.d.ts
TokenLocking.d.ts
index.d.ts
```

the contracts that are not versioned are put into the main folder while the contracts that are versioned have folders with version number subfolders:

```
├── CoinMachine
│   ├── 1
│   │   ├── CoinMachine.d.ts
│   │   ├── factories
│   │   │   └── CoinMachine__factory.ts
│   │   └── index.d.ts
│   ├── 2
│   │   ├── CoinMachine.d.ts
│   │   ├── factories
│   │   │   └── CoinMachine__factory.ts
│   │   └── index.d.ts
│   └── 3
│       ├── CoinMachine.d.ts
│       ├── factories
│       │   └── CoinMachine__factory.ts
│       └── index.d.ts
```

The `clients` directory now looks like this:

```
├── ColonyNetworkClient.ts
├── Core
│   ├── ColonyClientV1.ts
│   ├── ColonyClientV2.ts
│   ├── ColonyClientV3.ts
│   ├── ColonyClientV4.ts
│   ├── ColonyClientV5.ts
│   ├── ColonyClientV6.ts
│   ├── ColonyClientV7.ts
│   ├── ColonyClientV8.ts
│   ├── ColonyClientV9.ts
│   ├── ColonyVersionClient.ts
│   ├── extensions
│   │   ├── commonExtensions.ts
│   │   ├── extensionsV3.ts
│   │   ├── extensionsV4.ts
│   │   ├── extensionsV5.ts
│   │   ├── extensionsV6.ts
│   │   ├── extensionsV7.ts
│   │   ├── extensionsV8.ts
│   │   ├── extensionsV9.ts
│   │   └── SetPaymentDomain.ts
│   └── __tests__
│       └── ColonyClientV1.ts
├── Extensions
│   ├── CoinMachine
│   │   ├── 1
│   │   │   ├── CoinMachineClientAddons.ts
│   │   │   └── CoinMachineClient.ts
│   │   ├── 2
│   │   │   ├── CoinMachineClientAddons.ts
│   │   │   └── CoinMachineClient.ts
│   │   └── 3
│   │       ├── CoinMachineClientAddons.ts
│   │       └── CoinMachineClient.ts
[...]

PR #510, opened 2022-03-03
Upgrade ethers to version 5

chmanie:
## Description

This PR upgrades ethers to v5 and adjustes the typechain target correspondingly.
A lot had to be touched here, especially because typechain for ethers 5
does not create simplified versions of the solidity overloads anymore,
so the critical overloads have been extracted into own augments.

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Updated `ethers` to v5
* Using `@typechain/ethers-v5` now

NOTE: this PR is based on #509, so please wait with a review until that is merged.


rdig (approved):
This includes all the abi changes as well because it was not rebased, right ?

chmanie:
Yes, I rebased it now

PR #513, opened 2022-03-10
Fix/refactor extension client types

chmanie:
## Description

This PR repairs the extension types, basically doing the same we're already doing for the Colony augments. It's a lot of refactoring, but nothing really noteworthy as we're only using already established patterns.

**Changes** 🏗

* Complete refactoring of extension augments.


rdig (approved):
All seems good here. Very nice! :+1:

PR #517, opened 2022-03-17
Feat/motion proofs

chmanie:
## Description

This PR adds proper permission proofs to the new `createMotion` method and uses the reputation oracle to determine the reputation values for the already existing proofs, removing the necessity to pass them in on the public API level.


rdig (approved):
These are some very nice and welcomed improvements.

PR #522, opened 2022-04-04
Upgrade typechain to version 10

chmanie:
## Description

This PR upgrades `typechain` and `@typechain/ethers-v5` to version 10.0.0 and 8.0.0 respectively. This required some changes in the way the ABI files were named and sorted into folders.

`__develop__` -> `latest`
`__static__` -> `__fixed__`

**Changes** 🏗

- `typechain`: `v10.0.0`
- `@typechain/ethers-v5`: `v8.0.0`
- Change ABI naming and regenerate contract types with new typechain versions. 

NOTE: this PR is based on #521. Please merge that one before reviewing this.

PR #523, opened 2022-04-04
Better package exports

chmanie:
## Description

This PR adds better package exports for esm and commonjs, de-duplicates the types to reduce package size and adds an `extras` module for optional use.

NOTE: This PR is based on #522, so please merge that one before reviewing this.


rdig (approved):
Looks good

PR #542, opened 2023-01-30
Fix/token exports

chmanie:
## Description

This PR moves the `TokenClient`, `TokenAuthority` and `TokenLocking` to the `token` endpoint (where they belong).

On the way I sorted out the difference between `constants` and `types` and moved things around accordingly.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@colony/colony-js",
-  "version": "6.1.0",
+  "version": "6.2.0-beta.1",
```
thomassdk, line 3: Is this a official release to `6.2.0`?

thomassdk (approved):
LGTM! Thanks for you work @chmanie my only question is about the release version in the package.json but not sure how what the release pipeline is for colonyJS

package.json
chmanie: The release will be done and adjusted accordingly when this is merged.

PR #567, opened 2023-07-25
Upgrade ethers to version 6

chmanie:


chmanie:
Not going to happen. We'll switch to wagmi/viem/abitype at some point.

PR #593, opened 2024-01-22
fix: correct ipfs metadata validation

chmanie:
## Description

Fixes the validation schema of ipfs metadata


rdig (approved):
Thank you for you fix!

PR #611, opened 2024-07-15
Add USDC token addresses and networkIds

chmanie:
## Description

This adds USDC token addresses and an enum with all supported network ids.


jakubcolony (approved):
This looks good, thanks @chmanie 👍

